Issue #7, opened 2022-08-16
vts needs configurable options


dreemkiller:
Currently, the `config.Store` options in `vts/cmd/main.go` for `PluginManagerCfg`, `TaStoreCfg`, `EnStoreCfg`, and `VtsGrpcConfig` are hard-coded. 

There should be a way for users of the module to configure those without modifying the source code.

PR #14, opened 2022-08-22
Read VTS config from YAML/JSON

setrofim:
Impalement a config reader to read in Store settings form YAML. Use it to
configure VTS.

This addresses https://github.com/veraison/services/issues/7


yogeshbdeshpande:
@setrofim As this is touching both the Provisioning and Verification use case, request please do an end to end test as well, just to ensure that everything is still working intact with both P and V pipelines.

config/ireader.go
```diff
@@ -0,0 +1,33 @@
...
+
+	// GetStores returns a map of Stores parsed using Read or ReadFile,
+	// keyed by the names under which they appeared in the configuration.
+	GetStores() map[string]Store
+
+	// GetStore returns a Store that was loaded under the specified name.
+	// If no such Store exists, an error is returned.
+	GetStore(name string) (Store, error)
+
+	// MustGetStore returns a Store that was loaded under the specfied
```
yogeshbdeshpande, line 30:
```suggestion
	// MustGetStore returns a Store that was loaded under the specified
```

thomas-fossati (approved):
excellent stuff.

though we could also use [viper](https://github.com/spf13/viper) :-)

config/ireader.go
```diff
@@ -0,0 +1,33 @@
...
+	Read (buf []byte) (int, error)
+
+	// ReadFile reads configuration from a file located at the specified
+	// path, returning the total number of bytes read, and any errors
+	// encountered.
+	ReadFile(path string) (int, error)
+
+	// GetStores returns a map of Stores parsed using Read or ReadFile,
+	// keyed by the names under which they appeared in the configuration.
+	GetStores() map[string]Store
```
yogeshbdeshpande, line 24: Should we use some other word for Store, as if I am not mistaken is a generic config container?

As this coincides with KVStore and Store in general meaning to TA or  RefVal Stores...

yogeshbdeshpande (approved):
LGTM! Few .nits included!

setrofim:
> though we could also use [viper](https://github.com/spf13/viper) :-)

True. I figured I'd keep things simple for now. We can always add viper as an IReader implementation once we require more sophisticated config management.

PR #18, opened 2022-08-23
Correct Script to include both PSA Profiles

yogeshbdeshpande:


thomas-fossati (commented):
I don't understand the need for this change: PSA endorsements only have one profile defined: https://www.ietf.org/archive/id/draft-fdb-rats-psa-endorsements-01.html#section-3.1-2

yogeshbdeshpande:
As discussed we no longer need the PR as PSA Endorsement Profile is different than Evidence Profile!

PR #23, opened 2022-08-30
mk: check for mockgen before test

setrofim:
Make sure mockgen is installed on the system before attempting to run
tests, requesting that the user install it if it isn't.

### DIFF
--- a/mk/test.mk
+++ b/mk/test.mk
@@ -25,11 +25,21 @@ THIS_DIR := $(dir $(abspath $(lastword $(MAKEFILE_LIST))))
 _mocks: $(MOCK_FILES)
 .PHONY: _mocks
 
-test: test-hook-pre realtest
+test: check-mockgen test-hook-pre realtest
 .PHONY: test
 
 test-hook-pre:
 .PHONY: test-hook-pre
+
+check-mockgen:
+	@if [[ ! -x $(MOCKGEN) ]]; then \
+		echo ""; \
+		echo "please install mockgen using:"; \
+		echo ""; \
+		echo "    go install github.com/golang/mock/mockgen@1.6.0"; \
+		echo ""; \
+	fi
+.PHONY: check-mockgen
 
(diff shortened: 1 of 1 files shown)


mk/test.mk
```diff
@@ -25,12 +25,22 @@ THIS_DIR := $(dir $(abspath $(lastword $(MAKEFILE_LIST))))
...
 
 test-hook-pre:
 .PHONY: test-hook-pre
 
+check-mockgen:
+	@if [[ ! -x $(MOCKGEN) ]]; then \
+		echo ""; \
+		echo "please install mockgen using:"; \
+		echo ""; \
+		echo "    go install github.com/golang/mock/mockgen@latest"; \
```
yogeshbdeshpande, line 39: We should be consistent with CI Files, where we request to install 1.6.0 version of mockgen?

yogeshbdeshpande (approved):

yogeshbdeshpande:
Just one question:

(a) Either we remove the mockgen if we are checking all the autogen files and mock folder completely in github?
and
remove the mockgen step completely from make files?

OR

(b) We need to do similar exercise for protoc and other autogen as well?

thomas-fossati:
> Just one question:
> 
> (a) Either we remove the mockgen if we are checking all the autogen files and mock folder completely in github? and remove the mockgen step completely from make files?

even if we check all the generated files in we still need to have the mockgen rules in place.

thomas-fossati (approved):
Another (functionally equivalent) way to achieve the same is to group all tools we depend on in a `tools.mk` like this: https://github.com/ietf-rats/ietf-corim-cddl/blob/main/tools.mk

setrofim:
> Another (functionally equivalent) way to achieve the same is to group all tools we depend on in a tools.mk like this: https://github.com/ietf-rats/ietf-corim-cddl/blob/main/tools.mk

This is worth considering if there are more such tools in the future. For now, given that we only need the check for `mockgen`, and that we only want to check for it on test, IMO it makes sense to have this inside `test.mk`

Issue #25, opened 2022-09-06
BUG: vts policymanager getPolicy will never return ErrNoPolicy


dreemkiller:
### What version of the package are you using?
aa55d1502134308f1f3d66399ca5a8bd378b47c2

### Does this issue reproduce with the latest release?
Yes (that's latest)

### What OS and CPU architecture are you using (`go env`)?
```
GO111MODULE=""
GOARCH="amd64"
GOBIN=""
GOCACHE="/home/ec2-user/.cache/go-build"
GOENV="/home/ec2-user/.config/go/env"
GOEXE=""
GOEXPERIMENT=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GOINSECURE=""
GOMODCACHE="/home/ec2-user/go/pkg/mod"
GONOPROXY=""
GONOSUMDB=""
GOOS="linux"
GOPATH="/home/ec2-user/go"
GOPRIVATE=""
GOPROXY="https://proxy.golang.org,direct"
GOROOT="/usr/local/go"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GOTOOLDIR="/usr/local/go/pkg/tool/linux_amd64"
GOVCS=""
GOVERSION="go1.19"
GCCGO="gccgo"
GOAMD64="v1"
AR="ar"
CC="gcc"
CXX="g++"
CGO_ENABLED="1"
GOMOD="/home/ec2-user/VeracruzVerifier/go.mod"
GOWORK=""
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -Wl,--no-gc-sections -fmessage-length=0 -fdebug-prefix-map=/tmp/go-build3596495490=/tmp/go-build -gno-record-gcc-switches"
```

### What did you do?

It's complicated, since I'm calling it from Veracruz. I'm using the kvstore/Memory backend.

### What did you expect to see?
`getPolicy` should return `ErrNoPolicy` when there is no policy loaded

### What did you see instead?
`getPolicy` returns the error generated by `o.Store.Get`, which is not `ErrNoPolicy`
The error generated is from `kvstore/memory.go`, line 49 in my version, of the format:
```
return nil, fmt.Errorf("key %q not found", key)
```
[...]

PR #26, opened 2022-09-06
getPolicy should return `ErrNoPolicy` when `o.Store.Get` fails

dreemkiller:
This is a fix for #25

### DIFF
--- a/vts/policymanager/policymanager.go
+++ b/vts/policymanager/policymanager.go
@@ -67,7 +67,7 @@ func (o *PolicyManager) getPolicy(ev *proto.EvidenceContext) (*policy.Policy, er
 
 	vals, err := o.Store.Get(policyID)
 	if err != nil {
-		return nil, err
+		return nil, ErrNoPolicy
 	}
 
 	// TODO(setrofim): for now, assuming that there should be exactly one


dreemkiller:
It's possible that we also want to remove:
```
	if len(vals) == 0 {
		return nil, ErrNoPolicy
```
below this change, as `o.Store.Get` will never return a `vals` with length `0`

vts/policymanager/policymanager.go
```diff
@@ -67,7 +67,7 @@ func (o *PolicyManager) getPolicy(ev *proto.EvidenceContext) (*policy.Policy, er
 
 	vals, err := o.Store.Get(policyID)
 	if err != nil {
-		return nil, err
+		return nil, ErrNoPolicy
```
thomas-fossati, line 70: the only problem is we lose information about the root cause.

what about:
```suggestion
		return nil, fmt.Errorf("%w: %v", ErrNoPolicy, err)
```

And if the caller needs to branch based on the error type they still can do that:
```
        if errors.Is(err, ErrNoPolicy) {
                // specific handling for "no policy"
        } else {
                // generic error handling
        }
```

dreemkiller: That should work

setrofim (changes requested):
Unfortunately, this solution won't work as `ErrNoPolicy`  is meant expressly to indicate that no error occurred in retrieval, but the policy doesn't exist. It gets "swallowed" here: https://github.com/veraison/services/blob/main/vts/policymanager/policymanager.go#L44

If any error other than "not found" occurs, we want to propagate it, not swallow it. So we don't want to wrap all errors returned from `Get()` as `ErrNotFound`.

I made the wrong assumption about KVStore API. I think the correct fix might be to push this down into KVStore, and have `IKVStore.Get()` return `ErrKeyDoesNotExist` or something.

I'll look into it.

dreemkiller:
I was afraid that modifying `IKVStore.Get()` might be the best answer.

yogeshbdeshpande:
> Unfortunately, this solution won't work as `ErrNoPolicy` is meant expressly to indicate that no error occurred in retrieval, but the policy doesn't exist. It gets "swallowed" here: https://github.com/veraison/services/blob/main/vts/policymanager/policymanager.go#L44
> 
> If any error other than "not found" occurs, we want to propagate it, not swallow it. So we don't want to wrap all errors returned from `Get()` as `ErrNotFound`.
> 
> I made the wrong assumption about KVStore API. I think the correct fix might be to push this down into KVStore, and have `IKVStore.Get()` return `ErrKeyDoesNotExist` or something.
> 
> I'll look into it.

We should raise a new github issue to fix it properly in KVStore and close this PR referring to it, then!

setrofim:
This is superseded by https://github.com/veraison/services/pull/30

PR #36, opened 2022-09-13
Enhance Policy Manager Unit Tests

yogeshbdeshpande:


policy/agent.go
```diff
@@ -42,97 +35,7 @@ func CreateAgent(cfg config.Store) (*PolicyAgent, error) {
...
-	evaluatedResult.AppraisalPolicyID = policy.ID
-
-	return &evaluatedResult, nil
-}
-
-func (o *PolicyAgent) Close() {
-	o.Backend.Close()
+func NewAgent(b IBackend) IAgent {
+	policy := &PolicyAgent{Backend: b}
+	return policy
```
thomas-fossati, line 40: This method seems redundant.  It's part of the public interface but it is never going to be used as-is (IIUC).  I think we can drop it and just
```go
    return &PolicyAgent{Backend: backend}
```
at L:35

( and maybe `CreateAgent` should be `NewAgent` )

policy/ibackend.go
```diff
@@ -9,7 +9,7 @@ import (
 type IBackend interface {
 	Init(cfg config.Store) error
 	GetName() string
-	Evaluate(
+	BackEndEvaluate(
```
thomas-fossati, line 12: why do we need to change this function name?

policy/iagent.go
```diff
@@ -0,0 +1,21 @@
...
+import (
+	"context"
+
+	"github.com/veraison/services/config"
+	"github.com/veraison/services/proto"
+)
+
+type IAgent interface {
+	SetBackEnd(b IBackend)
+	GetBackEnd() IBackend
```
thomas-fossati, line 12: inconsistent camelcase (should be `GetBackend`, `SetBackend`)

but more importantly: why are these two methods needed?  I thought the agent is created once and for all from config, whereas here we'd allow callers to swap it on the fly.

policy/agent.go
yogeshbdeshpande: Fair enough. I can change this ..

policy/ibackend.go
yogeshbdeshpande: I thought it is better to have different interface name in the same package.

thomas-fossati: I'd rather `ibackend::evaluate` than `ibackend::backend_evaluate`, no?

policy/agent.go
yogeshbdeshpande: I will remove NewAgent() to inline in CreateAgent, but will leave CreateAgent() name unchanged as it is more Verbose in a sense we are creating a Policy Agent using the specified backed

policy/ibackend.go
yogeshbdeshpande: OK, I am not too fussy about names, I can revert it back to Evaluate, if that is more convenient!

policy/iagent.go
yogeshbdeshpande: To your question in relevance of SetBackend: It would be a nice to have feature to allow callers to have a control about what BackEnd they want to use ? I am not certainly using this feature as of now, but can keep it if we agree!

For GetBackend, we are actually using it to check the Backend in the tests and now it been an interface, I do not see any wrong in keeping this so would like to retain this...

thomas-fossati: > To your question in relevance of SetBackend: It would be a nice to have feature to allow callers to have a control about what BackEnd they want to use ? I am not certainly using this feature as of now, but can keep it if we agree!

I think the intended API contract is to create the object from config and then use it with the understanding that it won't change midlife.
Therefore, I don't think setter is needed (or even wanted).

> For GetBackend, we are actually using it to check the Backend in the tests and now it been an interface, I do not see any wrong in keeping this so would like to retain this...

If it is only used in tests, then it's not necessary to put it in the public interface: since tests are in the same package as the implementation, they can access the object attributes directly.

In summary, I suggest we drop both setter and getter.

policy/agent.go
thomas-fossati: I don't get the rationale but I am OK with your decision :-)

(Mine was just a proposal to align with usual go conventions.)

policy/iagent.go
yogeshbdeshpande: We cannot check from an Interface, which Backend is been used without this method in the test?

thomas-fossati: ah right, the Create does not return a concrete type that can be inspected directly, OK!  Still it doesn't feel right to expose an interface method only for testing reasons.

yogeshbdeshpande: The change is inline with the current API GetBackendname().

I see no issue with allowing external entities to query what is set from Config, which is Name of the Backend and Backend itself??

PR #37, opened 2022-09-14
Align trust vector with draft-ietf-rats-ar4si

setrofim:
Bring the TrustVector within Attestation result into closer alignment
with https://datatracker.ietf.org/doc/draft-ietf-rats-ar4si/

- Align the claims inside the TrustVector with those specified by the
  ar4si spec.
- Change their values to be integers with values in the int8 range
  (note: due to protobuf limitations, actual representation is int32).
- Add a TrustTier type that corresponds to the "Trustworthiness Tier"
  concept from ar4si. ARStatus has methods to covert its value into a
  tier.
- The overall Test status is now a TrustTier. This is now set
  automatically by the core verifier from the TrustVector, rather than
  relying on Scheme plugins to update it (note: plugins can still
  override it, if necessary).
- Add "veraison-verifier-added-claims" extension to the attestation
  result and allow it to be populated by policy.


vts/plugins/scheme-psa-iot/main.go
```diff
@@ -64,7 +64,10 @@ func (s Scheme) GetFormat() proto.AttestationFormat {
 	return proto.AttestationFormat_PSA_IOT
 }
 
-func (s Scheme) SynthKeysFromSwComponent(tenantID string, swComp *proto.Endorsement) ([]string, error) {
+func (s Scheme) SynthKeysFromSwComponent(
```
yogeshbdeshpande, line 67: There seems to be no change except the newlines introduced. Any specific reason for this change?

vts/plugins/scheme-psa-iot/main.go
```diff
@@ -136,7 +139,10 @@ func (s Scheme) GetTrustAnchorID(token *proto.AttestationToken) (string, error)
 	), nil
 }
 
-func (s Scheme) ExtractVerifiedClaims(token *proto.AttestationToken, trustAnchor string) (*scheme.ExtractedClaims, error) {
+func (s Scheme) ExtractVerifiedClaims(
```
yogeshbdeshpande, line 142: Same question as line 67

proto/result.go
```diff
@@ -0,0 +1,434 @@
...
+import (
+	"fmt"
+
+	timestamppb "google.golang.org/protobuf/types/known/timestamppb"
+)
+
+// Claim values as defined by
+// https://datatracker.ietf.org/doc/draft-ietf-rats-ar4si/
+const (
+	// common
```
yogeshbdeshpande, line 14: What does //common here means?

vts/plugins/scheme-psa-iot/main.go
setrofim: To keep the lines short, to prevent wrapping and make them easier to parse.

proto/result.go
setrofim: The group of consts bellow are common to all claims. (The subsequent groups are titled with the claim they're for).

proto/result.go
```diff
@@ -0,0 +1,434 @@
...
+	// interpretation is claim-dependent. As such, I decided to preserve
+	// this by defining separate consts for each claim, along with
+	// associated explanatory comments.
+	// Contrast this with the INVALID value "99", which, though
+	// separately for each claim, has identical interpretations for all of
+	// them.
+
+	// configuration:  A Verifier has appraised an Attester's configuration,
+	//and is able to make conclusions regarding the exposure of known
+	//vulnerabilities.
```
thomas-fossati, line 44:
```suggestion
	// and is able to make conclusions regarding the exposure of known
	// vulnerabilities.
```

proto/result.go
```diff
@@ -0,0 +1,434 @@
...
+
+import (
+	"fmt"
+
+	timestamppb "google.golang.org/protobuf/types/known/timestamppb"
+)
+
+// Claim values as defined by
+// https://datatracker.ietf.org/doc/draft-ietf-rats-ar4si/
+const (
```
thomas-fossati, line 13: this is awesome, and it's key information for plugin writers.  I am wondering how to make it as visible as possible?

thomas-fossati (approved):
🏆

yogeshbdeshpande (approved):

yogeshbdeshpande:
Just a side note:

1) Perhaps we should create a github issue to move Attestation Results to a new repo, as discussed in our meeting.

2) This change adds a change in apiclient and evcli tools as well, so we should track that effort as well!

thomas-fossati:
> 2. This change adds a change in apiclient and evcli tools as well, so we should track that effort as well!

which change are you thinking of?

yogeshbdeshpande:
> > 2. This change adds a change in apiclient and evcli tools as well, so we should track that effort as well!
> 
> which change are you thinking of?



> > 2. This change adds a change in apiclient and evcli tools as well, so we should track that effort as well!
> 
> which change are you thinking of?

I was mentioning about reading the Attestation Results and echoing what has been received in return response for Veraison. At the moment this bit of code is not present.

thomas-fossati:
> At the moment this bit of code is not present.

If that's the case there should be no change to track, right?  What am I missing?

yogeshbdeshpande:
> > At the moment this bit of code is not present.
> 
> If that's the case there should be no change to track, right? What am I missing?

Well, we have https://github.com/veraison/evcli/issues/12 on evcli -> I think we do not need anything here. So we are ok.
We do not have a similar issue on apiclient ? Need tracking there as well...

yogeshbdeshpande:
> > > At the moment this bit of code is not present.
> > 
> > 
> > If that's the case there should be no change to track, right? What am I missing?
> 
> Well, we have [veraison/evcli#12](https://github.com/veraison/evcli/issues/12) on evcli -> I think we do not need anything here. So we are ok. We do not have a similar issue on apiclient ? Need tracking there as well...

Discussed the confusion: https://github.com/veraison/evcli/issues/13 will address the core problem.

For apiclient, let the composer of the library display this information. Suitable api functions will be provided for users to include so that they have the required information at their disposal.

PR #43, opened 2022-09-30
Implement a CLI for policy management.

setrofim:
- Switch to using Viper for configuration.
 - Change `IKVStore.Del()` to return an error if the specified key does not exist.
 - Add `IKVStore.Setup()` that performs one-time setup of the store (e.g. creating SQL tables).
 - Add `IKVStore.GetKeys()` that returns all keys in the store.
 - Add `policy.Store` that wraps `IKVStore` and handles policy version management.
 - Add CLI interface to the policy store.


policy/agent.go
```diff
@@ -3,34 +3,128 @@
...
+
+	evalBytes, err := json.Marshal(resultMap)
+	if err != nil {
+		return nil, fmt.Errorf("could not marshal updated result: %w", err)
+	}
+
+	var evaluatedResult proto.AttestationResult
+
+	if err = evaluatedResult.UnmarshalJSON(evalBytes); err != nil {
+
```
thomas-fossati, line 115:
```suggestion
```

provisioning/cmd/main.go
```diff
@@ -21,26 +22,44 @@ var (
...
+	cfg.SetDefault("list-addr", DefaultListenAddr)
+
+	return cfg, nil
+}
+
+func main() {
+	cfg, err := initConfig()
 	if err != nil {
-		log.Fatalf("could not get listen-addr: %v", err)
+		log.Fatalf("counl not read config: %v", err)
```
thomas-fossati, line 55: typo: counl->could

verification/cmd/main.go
```diff
@@ -2,28 +2,46 @@ package main
...
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	VTSClientCfg.AddConfigPath(wd)
+	VTSClientCfg.SetConfigType("yaml")
+	VTSClientCfg.SetConfigName("config")
+
+	err = VTSClientCfg.ReadInConfig()
+	if _, ok := err.(viper.ConfigFileNotFoundError); ok {
```
thomas-fossati, line 33: I think `errors.Is(err, viper.ConfigFileNotFoundError)` should be the idiomatic way to write this

verification/cmd/main.go
```diff
@@ -2,28 +2,46 @@ package main
...
+		// If there is no config file, use the defaults sete above.
```
thomas-fossati, line 34: typo: sete->set

vts/cmd/main.go
```diff
@@ -9,49 +9,57 @@ import (
...
 
 func main() {
-	cfg := config.NewYAMLReader()
+	v := viper.New()
 
-	_, err := cfg.ReadFile("config.yaml")
+	wd, err := os.Getwd()
 	if err != nil {
-		log.Fatalf("counfig.yaml could not be read.")
+		log.Fatalf("%v", err)
```
thomas-fossati, line 26: if we are not adding any context to this, `log.Fatal(err)` is enough

thomas-fossati (approved):
LGTM

kvstore/ikvstore.go
```diff
@@ -2,31 +2,41 @@
 // SPDX-License-Identifier: Apache-2.0
 package kvstore
 
-import "github.com/veraison/services/config"
+import "github.com/setrofim/viper"
```
yogeshbdeshpande, line 5: Avoid personal github's but I am sure this is a Stop gap measure, please raise a guthub issue to remember to revert it back once you have the required fix!

policy/cmd/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+Add policy from a file under the specified ID:
+
+    ./polcli add opa://1 path/to/policy.rego
+
+(Note: ID must be in the form "opa://<tenant id>", where <tenant id> is the
+integer ID of the tenant for whom the policy will be added. The "opa://" prefix
+indicates the policy format; currently, only OPA rego policies are supported.)
+
+Update and existing ID with a new version (or add if ID doesn't already exist):
```
yogeshbdeshpande, line 26:
```suggestion
Update an existing ID with a new version (or add if ID doesn't already exist):
```

policy/store.go
```diff
@@ -0,0 +1,160 @@
...
+
+	newPolicyBytes, err := json.Marshal(newPolicy)
+	if err != nil {
+		return err
+	}
+
+	return o.KVStore.Add(id, string(newPolicyBytes))
+}
+
+// Get returns the slice of all Policies associated with he specfied ID. Each
```
yogeshbdeshpande, line 72:
```suggestion
// Get returns the slice of all Policies associated with the specified ID. Each
```

policy/store.go
```diff
@@ -0,0 +1,160 @@
...
+		}
+
+		policies = append(policies, p)
+	}
+
+	return policies, nil
+}
+
+// List returns []Policy containing latest versions of all policies. All
+// policies returned will have distict IDs. In cases where multiple policies
```
yogeshbdeshpande, line 98:
```suggestion
// policies returned will have distinct IDs. In cases where multiple policies
```

yogeshbdeshpande (approved):
/Nit: left few comments, for minor corrections! in general LGTM!

policy/cmd/cmd/add.go
```diff
@@ -0,0 +1,69 @@
+// Copyright 2022 Contributors to the Veraison project.
+// SPDX-License-Identifier: Apache-2.0
+package cmd
```
thomas-fossati, line 3: mmm `cmd/cmd` stutters a bit.  Is there a better way to structure the naming here?

thomas-fossati (approved):
another batch of comments

verification/cmd/main.go
setrofim: It would be, however `errors.Is` matches on text, which could be different depending on where this originates. It is more robust to much on type here.

kvstore/ikvstore.go
setrofim: Issue raised: https://github.com/veraison/services/issues/44

policy/cmd/cmd/add.go
setrofim: The use of `cmd` to contain `main.go` is consistent with the other executables in the services repo. The use of `cmd` subdirectory at the same level as `main.go` to contain individual command is the structure suggested by Cobra (https://github.com/spf13/cobra/blob/main/user_guide.md). Ergo `cmd/cmd/`.  But I agree, it _is_ a bit awkward. 

I could either 
- Deviate from Cobra suggestion and either move the commands to the  same level as `main.go`.
- Deviate from Cobra suggestion and use a different name for the commands subdirctory (what?).
- Deviate from the structure used in other places in the services repo and go with the more common golangstructure of using the executable name under `cmd/`; note  that would give us `cmd/polcli/cmd/`, which is _a little_ less awkward, by we still have the somewhat confusing duplication of "cmd".
- Do something else..?

Any preference?

verification/cmd/main.go
setrofim: Elaborating on the above comment that features a couple of inaccuracies:

> however `errors.Is` matches <s>on text</s>

Technically, `errors.Is` is little more complicated (see: https://cs.opensource.google/go/go/+/refs/tags/go1.19.1:src/errors/wrap.go;l=40) It matches:
- If the target `error` is Comparable and equals (`==`) to the tested `error` 
- If the tested `error` implements an `Is` method, and calling it on target returns `true`.

`viper.ConfigFileNotFoundError` does not implement `Is` so only the first check is applied. Structs are compared based on their corresponding field values. `viper.ConfigFileNotFoundError` contains only two (hidden) fields, both of which are strings: https://github.com/spf13/viper/blob/master/viper.go#L111. So this _effectively_ amounts to matching on text.


>  which could be different depending on where <s>this originates</s>.

The problem isn't so much with where the error originates but with the two fields that get populated. They are the name of the config file, and the locations that are searched for it. Matching the string containing the locations, in particular, is going to be very brittle.

The upshot is that checking based on type seems more robust here.

thomas-fossati: Ah, I see! Thanks for digging into this.  Would the snippet below work instead?

```go
var cfgFileNotFound viper.ConfigFileNotFoundError

if errors.As(err, &cfgFileNotFound) {
    // ...
}
```

policy/cmd/cmd/add.go
thomas-fossati: Yes you are right that it's at least consistent, which is _nearly_ always a good thing :-)

> * Do something else..?

What about introducing a new convention?

From: 
`services/${subsystem}/cmd/`

To:
`services/${subsystem}/${executable_name}/`

setrofim: Mmm, I think it's useful to have executables readily identified by being contained under `cmd/`, especially in a code base as heterogeneous as ours. So, IMO `services/${subsystem}/cmd/${executable_name}/` would be better (and aligned with golang standard practices). I could also rename the subcommands directory to `commands`. So instead of `policy/cmd/cmd/`, we could have `policy/cmd/polcli/commands/`. This does technically deviate from the Cobra suggestions, but they do look like suggestions, rather than best practices, and we still have the subcommands neatly  isolated. WDYT?

verification/cmd/main.go
setrofim: Yup, that works! Updated.

policy/cmd/cmd/add.go
thomas-fossati: I like it!

setrofim: Updated. I also applied the same convention to the  other executables. Please could you re-ack the pull if the new directory structure looks alright to you .

thomas-fossati (approved):
awesome, thanks!

Issue #46, opened 2022-10-18
feature: each service should have a versioned "is running" route


dreemkiller:
### Is your feature request related to a problem? Please describe

When starting a Veraison-based service as part of my test suite, I find I need to add `sleep` commands to wait for the service to be up. The amount of time required for this varies by platform, and has caused intermittent failures.

In specific, this problem is about the Provisioning service, but it's a general problem with all services.

In addition, in preparation for the support of multiple service versions, we have no simple way to detect which versions are supported by a service.

### If possible, describe a solution

If each service were to support a simple versioned `GET` path (for provisioning service, it might be to `/endorsement-provisioning/v1`), this could solve both problems. If a `GET` on a specific version returned non-success (`404`, I suppose), the caller would know that version was not supported by the service. If it returned "connection refused" the caller would know that the service wasn't up on that URL/port.

### Describe alternatives you have considered

We've attempted `sleep`, as mentioned above, but this adds unnecessary delays in our test suite and has shown intermittent failures on most platforms due to variable loads. The versioning question is also not solved with that solution.

thomas-fossati:
Thanks for raising this, it looks like a useful feature.

We could either:

* Support [`HEAD`](https://www.rfc-editor.org/rfc/rfc9110.html#section-9.3.2) on the existing API paths.
* Expose a [`.well-known`](https://www.rfc-editor.org/rfc/rfc8615) with metadata about the service(s) including high-level status indication that could be queried.

setrofim:
Implemented by https://github.com/veraison/services/pull/50

PR #50, opened 2022-10-26
Service State

setrofim:
Add `GET /status` handling for `provisioning` and `verification`.
 
This addresses https://github.com/veraison/services/issues/46 and partially addresses https://github.com/veraison/services/issues/51


provisioning/api/handler.go
```diff
@@ -50,6 +54,49 @@ const (
...
+	apiMediaTypeList, err := structpb.NewList([]interface{}{ProvisioningSessionMediaType})
+	if err != nil {
+		panic(err) // Should never happen as the value above is hard-coded.
+	}
+
+	decoderMediaTypeList, err := proto.NewStringList(o.DecoderManager.GetSupportedMediaTypes())
+	if err != nil {
+		ReportProblem(c,
+			http.StatusInternalServerError,
+			fmt.Sprintf("could not retrieve decoddr media types: %v", err),
```
thomas-fossati, line 80:
```suggestion
			fmt.Sprintf("could not retrieve decoder media types: %v", err),
```

thomas-fossati (approved):
LGTM!

provisioning/api/handler.go
```diff
@@ -50,6 +54,49 @@ const (
 	ProvisioningSessionMediaType = "application/vnd.veraison.provisioning-session+json"
 )
 
+func (o *Handler) GetServiceState(c *gin.Context) {
+	state := proto.ServiceState{
+		ServerVersion: "0.0.1",
```
yogeshbdeshpande, line 59: Should the Server Version be hard coded or should it be set from a Config file? Just a question:

provisioning/api/handler.go
yogeshbdeshpande, line 80:
```suggestion
			fmt.Sprintf("could not retrieve decoder media types: %v", err),
```

yogeshbdeshpande (approved):
LGTM! Left few questions but looks ok

provisioning/api/handler.go
thomas-fossati: I think version should be computed at build time from git metadata (tags, raw commit info) and assigned to a predefined global via `go build -w -X VERSION_VAR=VERSION_VAL`

thomas-fossati: Raised #51 to track this.

setrofim: No need :). Just pushed an update that sets versions from git.

yogeshbdeshpande: @setrofim Too cool 👍

thomas-fossati (approved):
awesome.

(In the commit msg we should note "Fix #46 and partially addresses #51")

Issue #49, opened 2022-10-25
Allow VTS to listen on multiple networks


SabreenKaur:
Currently the VTS backend listens for GRPC requests on one address and both provisioning and verification services communicate with VTS using this single address. Therefore currently VTS does not support listening on multiple networks. So instead we want the following, to address the complete separation of both provisioning and verification services:

- One network which contains the provisioning and VTS services (called network N1)
- One network which contains the verification and VTS services (called network N2)
- The ability for VTS to listen on both networks N1 and N2

PR #52, opened 2022-10-27
vts: allow listening on multiple interfaces.

setrofim:
Previously, VTS server used `vts.server-addr` configuration as the listening address. This, however, restricted to listening on a single network interface. Now, the server only uses the port form the setting, and listens on all system interfaces. For edge cases where restricting the interface is desired, `vts.listen-addr` setting is introduced which overrides `server-addr` if specified.

This addresses https://github.com/veraison/services/issues/49


vts/trustedservices/README.md
```diff
@@ -2,3 +2,9 @@
 
 - `server-addr` (optional): address of the VTS server in the form
   `<host>:<port>`. If not specified, this defaults to `127.0.0.1:50051`.
+  Unless `listen-addr` is specified (see below), VTS server will extract the
+  port to listen on from this setting (but will listen on all local interfaces)
+- `listen-addr` (optional): The address the VTS server will listen on in the
+  form `<host>:<port>`. Only specify this if you want to restrict the server to
+  listen on a particular interface; otherwise, the server will list on all
```
SabreenKaur, line 9: Perhaps here is it meant to say 'the server will _listen_ on all interfaces..."?

vts/trustedservices/README.md
```diff
@@ -2,3 +2,9 @@
 
 - `server-addr` (optional): address of the VTS server in the form
   `<host>:<port>`. If not specified, this defaults to `127.0.0.1:50051`.
+  Unless `listen-addr` is specified (see below), VTS server will extract the
+  port to listen on from this setting (but will listen on all local interfaces)
+- `listen-addr` (optional): The address the VTS server will listen on in the
+  form `<host>:<port>`. Only specify this if you want to restrict the server to
+  listen on a particular interface; otherwise, the server will listen on all
+  interfaces on the port specified in `server-addr`.
```
thomas-fossati, line 10: IIUC, the `host` part in `server-addr` is ignored?

If so, should we call it `server-port` instead?

setrofim: It is ignored by the server, but not the client.

setrofim: Well, the assumption is only there if they share the config file, in which case, it's unavoidable. If it's not the case, you modify the config file for each consumer as appropriate.

setrofim: Effectively, this means that in a non-distributed deployment, you do not need to create multiple config file -- all services can be pointed to the same one, while also keeping the structure of config simple (i.e. just having `vts` rather than needing `vts-server` and `vst-client` sections that may then require duplicate values). 

If the deployment is non-distributed, the the view of the address space is necessarily shared. If the deployment is distributed, then (barring  the config being on network storage) each service will have its own instance of config file so the values can be adjusted as appropriate.

thomas-fossati: just to be clear, in a split environment, is this configuration acceptable then:

* provisioning service
```
[...]
  server-addr: "vts.p-lb-pool.local:50051"
```

* verification service
```
[...]
  server-addr: "vts.v-lb-pool.local:50051"
```

* VTS
```
[...]
  listen-addr: ":50051"
```

setrofim: It is. Note that VTS could also alternatively be validly specified as
```
[...]
  server-addr: ":50051"
```

thomas-fossati (approved):

Issue #51, opened 2022-10-26
services build-time information


thomas-fossati:
We should make build-time information, including but not limited to version, available to all executables, both services and CLIs.

Re: version. One option would be for it to be computed at build time from git metadata (e.g., tags, commit hash, etc.) and assigned to a predefined global via `go build -w -X VERSION_VAR=VERSION_VAL` made available to all binaries.

setrofim:
Implemented by #50 and #55

thomas-fossati:
I think we should also have versioning in the two API frontend services.

PR #55, opened 2022-10-27
policy: polcli version set from git

setrofim:
Specify config.Version (which gets populated at build time by git commit/tag) as the version for the executable.

Together with  https://github.com/veraison/services/pull/50, this implements https://github.com/veraison/services/issues/51


thomas-fossati (approved):
🚢 it!

PR #91, opened 2022-11-21
Adding CCA Platform Verification Plugin

yogeshbdeshpande:


proto/vts.proto
```diff
@@ -38,7 +38,8 @@ enum EndorsementType {
 message Endorsement {
     AttestationFormat scheme = 1;
     EndorsementType type = 2;
-    google.protobuf.Struct attributes = 3;
+    uint32   subType = 3;
```
thomas-fossati, line 41: The naming convention used in this file uses underscores, so this should be `sub_type`.
Also, why uint?  This better be a string.
Besides, this should be `optional`?
Furthermore (English has so many synonyms!) we should add a comment here to document the "oddities" about this field, at least:
1) the fact that it is "public" but effectively opaque to Veraison
2) the fact that it assumes there is only ever going to be one single level of sub-type nesting
so that future-selves (as well as other Veraison users).

thomas-fossati: Additionally, field numbers must not be arbitrarily changed: see https://developers.google.com/protocol-buffers/docs/proto#assigning

yogeshbdeshpande: Few things:
Makes sense to correct the name to align with proto buffer naming conventions. Will do it!
Why uint: I thought we agreed that the subType is a numeric field, so that the use of numbers does not need to expose the exact sematic of what is on the interface?

Besides why optional: I think, keeping it mandatory does not hurt? It makes the interface consistent. Those plugins who do not want to use it can leave it, will have the default value on the interface ?

I can add comments, I have no issues with that!

Field Number not to be changed: That comment seems more for a finished product released in the field.

Ours is still in the development and debugging face and at the point where changing the sequence is not going to hurt us in any way! Otherwise adding it in the end makes it more difficult to comprehend!

thomas-fossati: > Why uint: I thought we agreed that the subType is a numeric field, so that the use of numbers does not need to expose the exact sematic of what is on the interface?

not sure how that would make a difference, with the disadvantage of making it less readable.

> Besides why optional: I think, keeping it mandatory does not hurt? It makes the interface consistent. Those plugins who do not want to use it can leave it, will have the default value on the interface ?

but that is the exact definition of optional, isn't it?

> Field Number not to be changed: That comment seems more for a finished product released in the field.
> 
> Ours is still in the development and debugging face and at the point where changing the sequence is not going to hurt us in any way! Otherwise adding it in the end makes it more difficult to comprehend!

it's not critical now but it's not good practice either.  Let's try and "do the right thing."

yogeshbdeshpande: > 
I am fine with using string for the semantics if we all agree.

For field numbering:

For numbering the field here the right thing is to define it next to type and number it sequentially as we are in development phase, rather than in the end which is difficult for readers to comprehend and connect with the main type as it hinders readability as well..

So I am all for changing the numbers, as I said earlier, it is still in the development phase and it is perfectly fine to modify the interface at this stage

thomas-fossati: > For field numbering:
> 
> For numbering the field here the right thing is to define it next to type and number it sequentially as we are in development phase, rather than in the end which is difficult for readers to comprehend and connect with the main type as it hinders readability as well..

how so?  numbers are for the wire, not for the reader.  The reader reads the field names.

yogeshbdeshpande: > > For field numbering:
> > For numbering the field here the right thing is to define it next to type and number it sequentially as we are in development phase, rather than in the end which is difficult for readers to comprehend and connect with the main type as it hinders readability as well..
> 
> how so? numbers are for the wire, not for the reader. The reader reads the field names.

I meant the interface in .proto should be logical and readers should be able to comprehend it properly.
Keeping type and sub_type together enables that readability.

Also regarding optionality : Please see: https://github.com/protocolbuffers/protobuf/issues/2497

thomas-fossati: > I meant the interface in .proto should be logical and readers should be able to comprehend it properly. Keeping type and sub_type together enables that readability.

Position is orthogonal to numbering.
 
> Also regarding optionality : Please see: [protocolbuffers/protobuf#2497](https://github.com/protocolbuffers/protobuf/issues/2497)

For proto3 the features are here: https://developers.google.com/protocol-buffers/docs/proto3#specifying_field_rules

yogeshbdeshpande: image

yogeshbdeshpande: Seems we cannot use optional as currently protoc-gen-go-json does not support optional fields

https://github.com/mitchellh/protoc-gen-go-json/issues/13

yogeshbdeshpande: Also verified, that if the field is not explicitly set, then the interface treats it as optional and omits the same on the other end of the interface!

thomas-fossati: There is a difference between `singular` and `optional` (cfr. the proto3 docs linked above)

proto/vts.proto
```diff
@@ -38,7 +38,10 @@ enum EndorsementType {
 message Endorsement {
     AttestationFormat scheme = 1;
     EndorsementType type = 2;
-    google.protobuf.Struct attributes = 3;
+    /* sub_type is used by schemes to classify range of 
+    Endorsement sub types for a given Endorsement type */
```
thomas-fossati, line 42: this is not the comment I meant.  What needs to be communicated is:

1. the fact that it is "public" but effectively opaque to Veraison
2. the fact that it assumes there is only ever going to be one single level of sub-type nesting

yogeshbdeshpande: For 1. We are saying that sub_type is used by schemes, is in effect saying that it is opaque to Veraison. Saying it is Opaque to Veraison will make it more  confusing as reader will not understand who is the user of this field. 

However, to make it more clear I will add is opaque to Veraison and used by schemes ....

2. This assumption is implicit as only one sub type field is their right? so can only have one usage, it is not an array nor this is an extensible type? Not sure what value add is achieved, by explicitly stating that ? However if you feel  it adds more clarity about it, I am happy to add this in comment

thomas-fossati: > For 1. We are saying that sub_type is used by schemes, is in effect saying that it is opaque to Veraison. Saying it is Opaque to Veraison will make it more confusing as reader will not understand who is the user of this field.

I don't see the confusion.  If it is opaque to Veraison, Veraison is not the user. 

> However, to make it more clear I will add is opaque to Veraison and used by schemes

So is it "more clear" or "more confusing"?  Now *I* am confused :-)

> 2. This assumption is implicit as only one sub type field is their right? so can only have one usage, it is not an array nor this is an extensible type? Not sure what value add is achieved, by explicitly stating that ? However if you feel  it adds more clarity about it, I am happy to add this in comment

My point is exactly that the implicit assumption needs to be made explicit.

yogeshbdeshpande: Done!

setrofim (approved):

go.mod
```diff
@@ -63,7 +64,8 @@ require (
 	github.com/lestrrat-go/httpcc v1.0.1 // indirect
 	github.com/lestrrat-go/httprc v1.0.4 // indirect
 	github.com/lestrrat-go/iter v1.0.2 // indirect
-	github.com/lestrrat-go/jwx v1.2.23 // indirect
+	github.com/lestrrat-go/jwx v1.2.25 // indirect
+	github.com/lestrrat-go/jwx/v2 v2.0.6 // indirect
```
yogeshbdeshpande, line 68: Within services, we should only use V2 ideally, we need to evaluate why v1 is still needed here??

go.mod
```diff
@@ -63,7 +64,8 @@ require (
 	github.com/lestrrat-go/httpcc v1.0.1 // indirect
 	github.com/lestrrat-go/httprc v1.0.4 // indirect
 	github.com/lestrrat-go/iter v1.0.2 // indirect
-	github.com/lestrrat-go/jwx v1.2.23 // indirect
+	github.com/lestrrat-go/jwx v1.2.25 // indirect
```
yogeshbdeshpande, line 67: Try go mod tidy and see if the multiple versions help improve the situation

proto/evidence.proto
```diff
@@ -4,9 +4,12 @@ package proto;
...
 
+
 message EvidenceContext {
 	string tenant_id = 1 [json_name = "tenant-id"];
 	string trust_anchor_id = 2 [json_name = "trust-anchor-id"];
-	string software_id = 3 [json_name = "software-id"];
+	string reference_id = 3 [json_name = "reference-id"];
 	google.protobuf.Struct evidence = 5;
+	// unprocessed evidence not handled by the Evidence Context
+	google.protobuf.Struct up_evidence = 6;
```
yogeshbdeshpande, line 14: Move this into issue, why do we need to unprocessed evidence?

provisioning/plugins/corim-psa-decoder/extractor.go
```diff
@@ -30,13 +31,14 @@ func (o *Extractor) SetProfile(p string) {
 // and to make ref attributes from them
 type MeasurementExtractor interface {
 	FromMeasurement(comid.Measurement) error
-	MakeRefAttrs(PSAClassAttributes) (*structpb.Struct, error)
+	GetRefValType() string
```
yogeshbdeshpande, line 34: Improve the comment a bit to explain the interface better

provisioning/plugins/corim-psa-decoder/ccaplatformconfigid.go
```diff
@@ -33,19 +33,23 @@ func (o *CCAPlatformConfigID) FromMeasurement(m comid.Measurement) error {
 	return nil
 }
 
-func (o CCAPlatformConfigID) MakeRefAttrs(c PSAClassAttributes) (*structpb.Struct, error) {
+func (o CCAPlatformConfigID) GetRefValType() string {
+	return "platform-config"
+}
+
+func (o CCAPlatformConfigID) MakeRefAttrs(c ClassAttributes, scheme string) (*structpb.Struct, error) {
```
yogeshbdeshpande, line 40: We do not scheme sctring here so just hard code to cca and use it to avoid confusion

scheme/ischeme.go
```diff
@@ -64,23 +64,28 @@ type IScheme interface {
...
+// XXX(yd) Thomas to answer your question: Here basically this is the key used
+// to fetch all the Endorsements (using reference value key), generated from claims
+// extracted from the token
 type ExtractedClaims struct {
-	ClaimsSet  map[string]interface{} `json:"claims-set"`
-	SoftwareID string                 `json:"software-id"`
+	ClaimsSet   map[string]interface{} `json:"claims-set"`
+	ReferenceID string                 `json:"reference-id"`
+	// UnprocessedClaimsSet are claims decoded by Scheme but not appraised
+	UnprocessedClaimsSet map[string]interface{} `json:"unprocessed-claims-set"`
```
yogeshbdeshpande, line 88: Remove this unprocessed for now!

vts/plugins/scheme-tpm-enacttrust/main.go
```diff
@@ -75,7 +75,7 @@ func (s Scheme) ExtractClaims(
 	trustAnchor string,
 ) (*scheme.ExtractedClaims, error) {
 	if token.MediaType != TPMEnactTrustTokenMediaType {
-		return nil, fmt.Errorf("wrong format: expect %q, but found %q",
+		return nil, fmt.Errorf("wrong mediaType: expect %q, but found %q",
```
yogeshbdeshpande, line 78: media type

vts/plugins/scheme-cca-ssd-platform/main.go
```diff
@@ -0,0 +1,459 @@
...
+	}
+
+	finalstr := ccaTaLookupKey(tenantID, implID, instID)
+	log.Printf("CCA Plugin TA CCA Look Up Key= %s\n", finalstr)
+	return []string{ccaTaLookupKey(tenantID, implID, instID)}, nil
+}
+
+func (s Scheme) GetSupportedMediaTypes() []string {
+	return []string{
+		"application/eat-cwt; profile=http://arm.com/CCA-SSD/1.0.0",
```
yogeshbdeshpande, line 132: application/eat-collection

This should be the new media type

vts/plugins/scheme-cca-ssd-platform/main.go
```diff
@@ -0,0 +1,459 @@
...
+
+	extracted.ReferenceID = ccaReferenceLookupKey(
+		token.TenantId,
+		MustImplIDString(ccaToken.PlatformClaims),
+	)
+	log.Printf("\n Extracted Reference ID Key = %s", extracted.ReferenceID)
+	return &extracted, nil
+}
+
+// ValidateEvidenceIntegrity, decodes CCA collection and then invokes Verify API of ccatoken library.
```
yogeshbdeshpande, line 185: remove full stop

vts/plugins/scheme-cca-ssd-platform/main.go
```diff
@@ -0,0 +1,459 @@
...
+// realm token.
+func (s Scheme) ValidateEvidenceIntegrity(
+	token *proto.AttestationToken,
+	trustAnchor string,
+	endorsementsStrings []string,
+) error {
+	var endorsement TaEndorsements
+
+	if err := json.Unmarshal([]byte(trustAnchor), &endorsement); err != nil {
+		log.Println("Could not decode Endorsements in ExtractVerifiedClaims")
```
yogeshbdeshpande, line 197: 197 no needed

vts/plugins/scheme-cca-ssd-platform/main.go
```diff
@@ -0,0 +1,459 @@
...
+		return fmt.Errorf("could not decode endorsement: %w", err)
+	}
+	ta := endorsement.Attr.VerifKey
+	block, rest := pem.Decode([]byte(ta))
+
+	if block == nil {
+		log.Println("could not get TA PEM Block during validating evidence integrity")
```
yogeshbdeshpande, line 204: remove 204

vts/plugins/scheme-cca-ssd-platform/main.go
```diff
@@ -0,0 +1,459 @@
...
+
+	// once the signature on the token is verified, we can claim the HW is
+	// authentic
+	result.TrustVector.Hardware = ear.GenuineHardwareClaim
+
+	swComps := filterRefVal(endorsements, "cca.sw-component")
+	match := matchSoftware(claims, swComps)
+	if match {
+		result.TrustVector.Executables = ear.ApprovedRuntimeClaim
+		log.Println("\n matchSoftware Success")
```
yogeshbdeshpande, line 305: repace  log.Println  to log.Debug

vts/plugins/scheme-cca-ssd-platform/main.go
```diff
@@ -0,0 +1,459 @@
...
+
+	platformConfig := filterRefVal(endorsements, "cca.platform-config")
+	match = matchPlatformConfig(claims, platformConfig)
+
+	if match {
+		result.TrustVector.Configuration = ear.ApprovedConfigClaim
+		log.Println("\n matchPlatformConfig Success")
+
+	} else {
+		result.TrustVector.Configuration = ear.UnsafeConfigClaim
```
yogeshbdeshpande, line 320: AR4si does not have symantic

provisioning/plugins/corim-psa-decoder/ccaplatformconfigid.go
yogeshbdeshpande: Agreed to add a comment and clarify why this is needed!

vts/plugins/scheme-tpm-enacttrust/main.go
```diff
@@ -50,7 +50,7 @@ func (s Scheme) SynthKeysFromTrustAnchor(tenantID string, ta *proto.Endorsement)
 
 func (s Scheme) GetTrustAnchorID(token *proto.AttestationToken) (string, error) {
 	if token.MediaType != TPMEnactTrustTokenMediaType {
-		return "", fmt.Errorf("wrong format: expect %q, but found %q",
+		return "", fmt.Errorf("wrong mediaType: expect %q, but found %q",
```
yogeshbdeshpande, line 53: correct media type

scheme/ischeme.go
yogeshbdeshpande: Done!

provisioning/plugins/corim-psa-decoder/ccaplatformconfigid.go
yogeshbdeshpande: Clarifying comment added!

provisioning/plugins/corim-psa-decoder/extractor.go
yogeshbdeshpande: Done!

go.mod
yogeshbdeshpande: Corrected all dependencies, now jwx only uses version v2, v2.0.8

go.mod
yogeshbdeshpande: Done

vts/plugins/scheme-cca-ssd-platform/main.go
yogeshbdeshpande: https://github.com/ietf-rats-wg/draft-ietf-rats-ar4si/issues/17
added to address this issue in ar4si draft

vts/plugins/scheme-cca-ssd-platform/main.go
```diff
@@ -95,7 +96,7 @@ func (s Scheme) SynthKeysFromRefValue(
 	}
 
 	finalstr := ccaReferenceLookupKey(tenantID, implID)
-	log.Printf("CCA Plugin CCA Reference Value Look Up Key= %s\n", finalstr)
+	log.Debug("CCA Plugin CCA Reference Value Look Up Key= %s\n", finalstr)
 
 	return []string{ccaReferenceLookupKey(tenantID, implID)}, nil
```
thomas-fossati, line 101: why not reusing `finalstr` instead of recomputing the value?

Besides, I suggest changing `finalstr` to something more meaningful, e.g.: `lookupKey`

vts/plugins/scheme-cca-ssd-platform/main.go
```diff
@@ -124,7 +125,7 @@ func (s Scheme) SynthKeysFromTrustAnchor(tenantID string, ta *proto.Endorsement)
 	}
 
 	finalstr := ccaTaLookupKey(tenantID, implID, instID)
```
thomas-fossati, line 127: ditto as above

vts/plugins/scheme-cca-ssd-platform/main.go
```diff
@@ -173,7 +174,7 @@ func (s Scheme) ExtractClaims(
 		token.TenantId,
 		MustImplIDString(ccaToken.PlatformClaims),
 	)
-	log.Printf("\n Extracted Reference ID Key = %s", extracted.ReferenceID)
+	log.Debug("\n Extracted Reference ID Key = %s", extracted.ReferenceID)
```
thomas-fossati, line 177: not sure why is there a `\n` at the beginning of the string?

vts/plugins/scheme-cca-ssd-platform/main.go
```diff
@@ -220,7 +221,7 @@ func (s Scheme) ValidateEvidenceIntegrity(
 	if err = ccaToken.Verify(pk); err != nil {
 		return err
 	}
-	log.Println("\n CCA platform token signature, realm token signature and cryptographic binding verified")
+	log.Debug("\n CCA platform token signature, realm token signature and cryptographic binding verified")
```
thomas-fossati, line 224: ditto (multiple times, see also below)

vts/plugins/scheme-cca-ssd-platform/main.go
```diff
@@ -344,7 +345,7 @@ func matchSoftware(evidence psatoken.IClaims, endorsements []Endorsements) bool
 		matched = true
 		var attr SwAttr
 		if err := json.Unmarshal(endorsement.Attr, &attr); err != nil {
-			log.Println("Could not decode sw attributes from endorsements")
+			log.Debug("Could not decode sw attributes from endorsements")
```
thomas-fossati, line 348: we should log the error too

vts/plugins/scheme-cca-ssd-platform/main.go
```diff
@@ -355,7 +356,7 @@ func matchSoftware(evidence psatoken.IClaims, endorsements []Endorsements) bool
 			break
 		}
 
-		log.Printf("MeasType Evidence: %s, Endorsement: %s", *evComp.MeasurementType, attr.MeasType)
+		log.Debug("MeasType Evidence: %s, Endorsement: %s", *evComp.MeasurementType, attr.MeasType)
```
thomas-fossati, line 359: pure quibbling: but having both `MeasType` and `MeasurementType` is quite puzzling :-)

vts/plugins/scheme-cca-ssd-platform/main.go
```diff
@@ -375,11 +376,11 @@ func matchPlatformConfig(evidence psatoken.IClaims, endorsements []Endorsements)
 		return false
 	}
 	if len(endorsements) > 1 {
-		log.Printf("matchPlatformConfig failed number of cca config %d > 1 ", len(endorsements))
+		log.Debug("matchPlatformConfig failed number of cca config %d > 1 ", len(endorsements))
```
thomas-fossati, line 379: I am unable to understand what we want to say with this message. Is it:
```golang
    log.Debug("got %d CCA configuration endorsements, want 1", len(endorsements))
```
If so, we seem to be constraining the number of possible configurations to one only.  Is this intended?  Shouldn't we cycle over all the existing until we find a match instead?

proto/vts.proto
```diff
@@ -37,7 +37,11 @@ enum EndorsementType {
 message Endorsement {
     string scheme = 1;
     EndorsementType type = 2;
-    google.protobuf.Struct attributes = 3;
+    /* sub_type is opaque to Veraison and is used by schemes to classify range of 
+    Endorsement sub types for a given Endorsement type. It is assumed that 
+    there is going to be only one single sub type required */
+    string   sub_type = 3;
```
thomas-fossati, line 43:
```suggestion
    string sub_type = 3;  
```

provisioning/plugins/corim-psa-decoder/ccaplatformconfigid.go
```diff
@@ -33,19 +33,26 @@ func (o *CCAPlatformConfigID) FromMeasurement(m comid.Measurement) error {
...
 }
 
-func (o CCAPlatformConfigID) MakeRefAttrs(c PSAClassAttributes) (*structpb.Struct, error) {
+func (o CCAPlatformConfigID) GetRefValType() string {
+	return "platform-config"
+}
+
+// MakeRefAttrs is an interface function to populate reference attributes.
+// For PlatformConfig scheme argument is not strictly required, but is required for other
+// usage of the same interface
```
thomas-fossati, line 42: this comment should go on the interface method rather than on an implementation of the interface.

provisioning/plugins/corim-psa-decoder/ccaplatformconfigid.go
```diff
@@ -33,19 +33,26 @@ func (o *CCAPlatformConfigID) FromMeasurement(m comid.Measurement) error {
...
+func (o CCAPlatformConfigID) MakeRefAttrs(c ClassAttributes, scheme string) (*structpb.Struct, error) {
```
thomas-fossati, line 43: I thought we agreed to rename `scheme` to `unused` and hardcode the required value in L45-47

yogeshbdeshpande: As far as I recollect we agreed to retain the usage but with a clarification that it is strictly not required.
This is more clear than using the hard coded one...

provisioning/plugins/corim-psa-decoder/ccaplatformconfigid.go
yogeshbdeshpande: I have tidied up the comment to only retain the relevant bits here and moved the top line comment to the interface.

PR #92, opened 2022-11-22
Switch to using github.com/veraison/ear

setrofim:
Use github.com/veraison/ear implementation of the Attestation Result
instead of a locally defined protobuf-based one.


proto/state.pb.go
```diff
@@ -1,7 +1,7 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
 // 	protoc-gen-go v1.27.1
-// 	protoc        v3.21.7
+// 	protoc        v3.21.9
```
yogeshbdeshpande, line 4: Do you really want to change the protoc version?

vts/appraisal/appraisal.go
```diff
@@ -0,0 +1,54 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+package appraisal
+
+import (
+	"github.com/veraison/ear"
+	"github.com/veraison/services/proto"
+)
+
+type Appraisal struct {
```
yogeshbdeshpande, line 11: I am struggling a bit to understand the use of Appraisal in context of:
May be its just me, but could you please add some comments to clarify how Appraisal is used in the flow ?
Especially given the fact that the interface is as below

func (s *RPCClient) AppraiseEvidence(ec *proto.EvidenceContext, endorsements []string) (*ear.AttestationResult, error) {

setrofim: It is the equivalent of `proto.AppraisalContext` but not tied to protobufs. It ties together the evidence and the attestation result. The difference is that it includes the actual attestation result and so can be used internally by VTS (specifically, by the policy manager). Since we're now using the EAR attestation result this is not defined as a protobuf message`, `proto.AppraisalContext` must now have it in a serialized form, which means it's no longer usable inside VTS, and is purely the definition of the protobuf message.

As requested, I've added a comment to the struct defintion.

proto/state.pb.go
setrofim: Yes. protoc is not in the repository, it's part of the environment. This is not repository specific, and we must constrain (let along dictate specific) tool version without a good reason. Not touching auto generated files that ought not be in the git repo in the first place is not a good reason.

yogeshbdeshpande: Different version of protoc triggers a modification of protoc files, which are not modified by the user in any !

We should avoid that. Isn't it?

setrofim: No. We should not have these files in the repo in the first place. But while we must, we have to accept that  innocuous changes may be added when they are regenerated. This is the lesser evil compared to forcing people to downgrade tools in their environment for no good reason.

vts/appraisal/appraisal.go
```diff
@@ -0,0 +1,57 @@
...
+package appraisal
+
+import (
+	"github.com/veraison/ear"
+	"github.com/veraison/services/proto"
+)
+
+// Appraisal provides an appraisal context internally within the VTS (e.g. for
+// policy evaluation). It is the analog of proto.AppraisalContext, but with a
+// ideserialized AttestationResult.
```
thomas-fossati, line 13: typo (extra 'i'): ideserialized

thomas-fossati (commented):
LGTM

proto/state.pb.go
yogeshbdeshpande: Yes, if we have a Docker build environment which every one uses,  then we will not have to commit the files, which the user has not ideally modified, like in this case! For now we have to live with this, the risk I see is, if any one using a slightly different (and also older version of protoc than what you are using here), will end up submitting these files again, with no real reason,

yogeshbdeshpande (approved):
LGTM!

proto/state.pb.go
setrofim: Well, there is no real risk here as the genes are purely in the comments. But in terms churn, then yes, this will keep happening unless everyone is on exactly same version of an external tool. Which outside a containerised build environment is just not tenable.

I also very much view the presence of these files  as a temporary thing (albeit with an undefined time limit). We _have_ to have them here at the moment as a workaround for the limitation of the go build system. However we must accept this as an inevitability. This is not the right way to do  thing, and it needs to be fixed. 

Which actually reminds me, while digging into something not directly related, one the libs I was looking it had some similar problem, and its solution was to have tests in a separate package.  I don't think was specifically due to protobufs, but working around basically the same problem of tests being unnecessarily hooked by go install.

verification/verifier/verifier.go
```diff
@@ -70,11 +70,8 @@ func (o *Verifier) ProcessEvidence(tenantID string, data []byte, mt string) ([]b
...
 		return nil, err
 	}
-	res, err := appraisalCtx.Result.MarshalJSON()
-	if err != nil {
-		return nil, err
-	}
-	fmt.Printf("Apprisal Context %s,", string(res))
-	return appraisalCtx.Result.MarshalJSON()
+
+	fmt.Printf("Apprisal Result: %s", string(appraisalCtx.Result))
```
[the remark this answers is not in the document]
setrofim: oh crap, yes. Well spotted!

policy/opa.go
```diff
@@ -166,20 +143,28 @@ func processUpdateValue(value interface{}) (map[string]interface{}, error) {
...
+			ErrBadOPAResult, value)
+		return nil, err
 	}
 
 	update := map[string]interface{}{
-		"status":                         updatedStatus,
-		"trust-vector":                   updateTv,
-		"veraison-verifier-added-claims": rawUpdate["veraison-verifier-added-claims"],
+		"ear.status":                         updatedStatus,
+		"ear.trustworthiness-vector":                   updateTv,
```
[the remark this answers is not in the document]
setrofim: One rabies short coming up!

Issue #79, opened 2022-11-16
feature: Store API: do not statically encode the attestation format


yogeshbdeshpande:
The attestation format must not be statically encoded in the protobuf definition because this forces all the attestation formats that a plugin may supply to be known at compile time.

setrofim:
Addressed by https://github.com/veraison/services/pull/100

PR #100, opened 2022-12-08
Remove AttestationFormat

setrofim:
We currently have three was of referencing attestation scheme specific functionality:

- media type
- attestation format
- attestation scheme name (implemented as string version of the above)

This change removes attestation format because:

- It is currently not being set properly (end-to-end logs show it as unset during verification path execution).
- It is not consistently used (which is why the above has gone unnoticed until now).
- It is made redundant by the media type and scheme name
- The enum defining valid values is hard-coded, meaning adding new plugins requires a modification to the core services code line.

In the situations where attestation format being used, the scheme name is used instead.

Note: currently, this means that the scheme name must be hard-coded in two places -- the scheme plugin, and the provisioning decoder plugin; and those two values must match. The upcoming plugin framework unification and restructuring will remove dependency and allow the name to be defined in a single place from which it will be used by both plugins.

This addresses https://github.com/veraison/services/issues/79

Signed-off-by: Sergei Trofimov <sergei.trofimov@arm.com>


thomas-fossati (approved):
LGTM

PR #115, opened 2023-02-06
Plugins

setrofim:
This is a major redesign of Veraison plugin handling framework. 

- It replaces provisioning and vts plugin managers with a single generics-based implementation. This allows plugins to be loaded form a single location, and to combine both plugins into a single executable. This also minimises functionality duplication and add a clean way to add further plugin types in the future.
- Scheme-specific code is moved out of `provisioning` and `vts` subtreets, and is collated in a single place under `schemes/`.
- A build configuration is added to remove plugin support from veraison services. In this configuration scheme handling is built into the service expecutables.

For instructions on migrating existing Veraison plugins see: https://github.com/veraison/services/blob/0182640982756bc9a361bc5037bd4d7282a93e29/scheme/MIGRATING.md


setrofim:
> **Note**: as previsouly discussed, there is one outstanding issue with the current refactior -- then naming of the VTS-side plugin interface. It has been renamed to `IEvidenceDecoder` since the name "scheme" isn't appropriate, given that both plugins can be equality thought to be part o fthe scheme. However, as noted, "evidence decoder" may not be the best name given the scop of its duties, however I'm having a hard time thinking of a better alternative. Related to this is the fact that it's name (accessed via `GetName` is referred to as a "scheme" by the endorsement decoder; this is not accurate (the "Scheme" in `Endorsement` struct refers to this plugin, not to the endorsement scheme).

decoder/README.md
```diff
@@ -0,0 +1,6 @@
+This package defines [`IEvidinceDecoder`](ievidecoder.go) and
+[`IEndorsmentDecoder`](iendorsementdocoder.go) [pluggable](../plugin/README.md)
```
thomas-fossati, line 2:
```suggestion
This package defines [`IEvidenceDecoder`](ievidencedecoder.go) and
[`IEndorsementDecoder`](iendorsementdocoder.go) [pluggable](../plugin/README.md)
```

decoder/README.md
```diff
@@ -0,0 +1,6 @@
...
+interfaces and associated RPC channels. These are used add new attestation
```
thomas-fossati, line 3:
```suggestion
interfaces and associated RPC channels. These are used to add new attestation
```

decoder/README.md
```diff
@@ -0,0 +1,6 @@
...
+scheme sort to Veraison services. Additionally, the package defines a [couple
+of wrappers](plugin.go) around `plugin.RegisterImplementation` for registering
+implementations of these two implementations.
```
thomas-fossati, line 6:
```suggestion
implementations of these two interfaces.
```

Makefile
```diff
@@ -14,4 +17,12 @@ SUBDIR += verification
 SUBDIR += vts
 SUBDIR += vtsclient
 
+COVERAGE_THRESHOLD := 60.0
+# plugin coverage is low because it is mostly tested via plugin/test, a
+# spearate package (this is necessary due to to the nature of the code being
```
yogeshbdeshpande, line 22:
```suggestion
# separate package (this is necessary due to to the nature of the code being
```

Makefile
```diff
@@ -14,4 +17,12 @@ SUBDIR += verification
...
+# tested. plugin/test coverage is lso because it's purely test code).
```
yogeshbdeshpande, line 23:
```suggestion
# tested. plugin/test coverage is low because it's purely test code).
```

plugin/README.md
```diff
@@ -0,0 +1,165 @@
+# Veraison Plugin Framework
```
thomas-fossati, line 1: awesome!

builtin/builtin_manager.go
```diff
@@ -0,0 +1,91 @@
+// Copyright 2023 Contributors to the Veraison project.
```
yogeshbdeshpande, line 1: Could you please also add a README.md to package builtin?
This way it is easy for reader to connect the dots with the rest of the sub-system

plugin/test/trooper/trooper.go
```diff
@@ -0,0 +1,28 @@
+// Copyright 2022-2023 Contributors to the Veraison project.
+// SPDX-License-Identifier: Apache-2.0
+package main
+
+import (
+	"github.com/veraison/services/plugin"
+	"github.com/veraison/services/plugin/test"
+)
+
+type ImperialTrooper struct {
```
thomas-fossati, line 10: LOL :-)

decoder/README.md
yogeshbdeshpande: > > Note: as previsouly discussed, there is one outstanding issue with the current refactior -- then naming of the VTS-side plugin interface. It has been renamed to `IEvidenceDecoder` since the name "scheme" isn't appropriate, given that both plugins can be equality thought to be part o fthe scheme. However, as noted, "evidence decoder" may not be the best name given the scop of its duties, however I'm having a hard time thinking of a better alternative. Related to this is the fact that it's name (accessed via `GetName` is referred to as a "scheme" by the endorsement decoder; this is not accurate (the "Scheme" in `Endorsement` struct refers to this plugin, not to the endorsement scheme).

Should we call IEvidenceDecoder to IEvidenceProcessor or IEvidenceHandler. They are atleast more generic then pure Decoding and renders a bit more clarity. WDYT ?

builtin/builtin_loader.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+
+	return nil
+}
+
+func GetBuiltinHandleByMediaType[I plugin.IPluggable](mediaType string) (I, error) {
+	return GetBuiltinHandleByMediaTypeUsing[I](defaultBuiltinLoader, mediaType)
+}
+
+func GetBuiltinHandleByMediaTypeUsing[I plugin.IPluggable](
```
yogeshbdeshpande, line 88: func GetBuiltinHandleByMediaTypeUsing[I plugin.IPluggable]( 
Did you mean to reword it with taking into consideration Using ??

scheme/MIGRATING.md
```diff
@@ -0,0 +1,85 @@
+This is a guide for converting plugins written for an earlier version of
+Veraison to the updated plugin framework.
+
+1. Remove `GetFormat()`. Attestation formats are no longer required --  the
+   combination of the plugin name and media types is now used to select plugins.
+2. Update `GetName()`. If this was previously implemented in terms of the
+   format. This can now be defined as a string constant.
+3. Replace `main()`. Instead of using HashiCorp APIs directly, Veraison no
```
yogeshbdeshpande, line 8:
```suggestion
3. Replace `main()`. Instead of using HashiCorp APIs directly, Veraison no longer
```

yogeshbdeshpande, line 8:
```suggestion
3. Replace `main()`. Instead of using HashiCorp APIs directly, Veraison now
```

scheme/psa-iot/extractor.go
```diff
@@ -113,7 +117,7 @@ func extractMeasurement(obj MeasurementExtractor, m comid.Measurement, class Cla
 		return &proto.Endorsement{}, fmt.Errorf("failed to create software component attributes: %w", err)
 	}
 	refVal := proto.Endorsement{
-		Scheme:     schemeName,
+		Scheme:     SchemeName,
```
yogeshbdeshpande, line 120: As discussed: this needs working through to resolve the correct schemeName as this part of the code is used by both PSA and CCA

decoder/README.md
setrofim: Yes, "handler", while still not perfect, makes more sense than "decoder". I'll make the change.

builtin/builtin_loader.go
setrofim: The "Using" is gontinued by the first paramer which lis the loader.  (see the call in line 85). I could say "UsingLoader", but that would just make an already long function longer without adding additional information.

setrofim:
I've addressed the review comments and made changes discussed with @yogeshbdeshpande: 

- "decoders" are now called "handlers", as their remit goes beyond merely decoding data.
-  `cca` scheme now has its own endorsement plugin, rather than piggibacking on the PSA one.
- added `GetAttestationScheme()` that returns the name of the attestation scheme implemented by a plugin (as opposed to the name of the actual plugin). This is now used by provisioning flow to identify the evidence handler plugin in order to synthesize store keys. This means there is no longer a non-obvious requirement for the evidence handler to be named after the scheme.

setrofim:
Also added additional `INFO` logging to `vts-service` and `provisioning-service` to report the media types the know about.

yogeshbdeshpande (approved):

yogeshbdeshpande:
Awesome work !

Thank you!

PR #117, opened 2023-02-09
Moved to latest version of CCA Token library

yogeshbdeshpande:
Signed-off-by: Yogesh Deshpande <yogesh.deshpande@arm.com>


yogeshbdeshpande:
@setrofim I have tested end to end setup with your inputs and corrected required CCA tests based on the new version of CCA token library.

setrofim (approved):
LGTM!

Though I think we should not be using development versions of `ccatoken`  and `psatoken` -- we should create releases and use those.

yogeshbdeshpande:
> LGTM!
> 
> Though I think we should not be using development versions of `ccatoken` and `psatoken` -- we should create releases and use those.

Yep, valid point, have addressed this more cleaner solution in then next version of go.mod

PR #121, opened 2023-02-22
EAR 2023 profile support

setrofim:
Update Entity Attestation Result support to the latest current version.


verification/verifier/iverifier.go
```diff
@@ -8,5 +8,5 @@ type IVerifier interface {
 	GetVTSState() (*proto.ServiceState, error)
 	IsSupportedMediaType(mt string) (bool, error)
 	SupportedMediaTypes() ([]string, error)
-	ProcessEvidence(tenantID string, data []byte, mt string) ([]byte, error)
+	ProcessEvidence(tenantID string, nonce []byte, data []byte, mt string) ([]byte, error)
```
thomas-fossati, line 11: Just a fly-by comment.  Since this is a generic interface (i.e., independent of the interaction model that is used to submit evidence) wouldn't it be better to avoid coupling with the challenge-response here?  Maybe we should pass a "verification session context" object instead?  Now, that'd contain a nonce for now because challenge-response is the only interface we expose, but in the future it may grow to support different things.  I'm happy with a Joe Strummeresque answer: "the future is unwritten" :-)

thomas-fossati (approved):
LGTM, thanks!

verification/verifier/iverifier.go
setrofim: I'd rather keep it simple and explicit for now. Introducing a session context would not reduce coupling. It'd be more generic, but also abstruse.

PR #122, opened 2023-02-22
scheme/{psa,cca}: add life cycle evaluation

setrofim:
Evaluate the security life cycle in order to populate the instance identity, runtime opaque, and storage opaque trust vector claims.


thomas-fossati (approved):
cool!

PR #129, opened 2023-03-20
README: de-authorize end-to-end section

setrofim:
Rephrase the end-to-end section of the README, to not imply that the end-to-end script is the authoritative way of doing provisioning and verification using Veraison, rather than just an example.


yogeshbdeshpande (approved):
Thanks! Yes, this clarifies it much better.
LGTM!

Issue #108, opened 2023-01-26
Align storage type for SW Component : Measurement Description  


yogeshbdeshpande:
The draft PSA Endorsements [draft-psa-endorsements](https://thomas-fossati.github.io/draft-psa-token/draft-tschofenig-rats-psa-token.html)
define:   ? &(measurement-desc: 6) => text

This is string identifying the hash algorithm used to compute the corresponding Measurement Value. 

However in the Endorsement database we store this as a uint64 number, corresponding to HashAlgID.

It is more convenient to align the storage with the string name corresponding to Algorithm Identifier.

This aligns the interface to the draft and enhances store readability when debugging as well.

PR #133, opened 2023-04-04
Align storage type for sw-comp measurement description

SabreenKaur:
This change modifies the type of the hash algorithm used to compute the measurement value, so that now it matches the type defined in the latest psa attestation token internet draft.

This addresses #108


setrofim (approved):
LGTM

Issue #109, opened 2023-01-27
BUG: Missing mandatory claim causes slient failure


SabreenKaur:
When missing a mandatory claim (e.g. multiple nonces specified) in the claims template that is used to generate the evidence used during verification, the application fails silently. There is no detail on the failure outputted in logging or any attestation result outputted. The application exits with a status: failed in the response body, when testing this using the Tavern testing framework.

The error happens when `scheme.GetTrustAnchorID(token)` is called in `services/vts/trustedservices_grpc.go`. An example of the error given when manually printed is the following:

```
Plugin.GetTrustAnchorID RPC call failed: failed CBOR decoding of PSA claims: decode failed for both p1 (validation of PSA claims failed: validating psa-client-id: missing mandatory claim) and p2 (validation of PSA claims failed: validating psa-nonce: wrong syntax for claim: got 2 nonces, want 1)
```

We expect to see the attestation result successfully output with the trust vector populated. For example, for a missing mandatory claim all entries in the trust vector should be populated with 1s

Issue #107, opened 2023-01-25
BUG: Unrecognised instance and implementation id in claims template causes silent failure


SabreenKaur:
When using an unrecognised psa-instance-id or psa-implementation-id in the claims template that is used to generate the evidence used during verification, the application fails silently. There is no detail on the failure outputted in logging or any attestation result outputted. The application exits with a `status: failed `in the response body, when testing this using the Tavern testing framework.

The failure happens inside the `getrustAnchor(id string)` function in `services/vts/trustedservices_grpc.go`. The  error happens when `TaStore.Get(id)` is called. An example of the error given when manually printed is the following:

```
key not found: "PSA_iot://0/YWNtZS1pbXBsZW1lbnRhdGlvbi1pZC0wMDAwMDAwMDE=/Aaq7zN0AESIzqrvM3QARIjOqu8zdABEiM6q7zN0AESIz
```
Where `Aaq7zN0AESIzqrvM3QARIjOqu8zdABEiM6q7zN0AESIz` is the invalid instance id used in the template that is used to synthesise the evidence.

We expect to see the attestation result successfully output with the trust vector populated. For example, for an unrecognised instance identity or implementation id the trust vector entry for the instance identity should be: `instance-identity:97`

setrofim:
Addressed by https://github.com/veraison/services/pull/135

PR #135, opened 2023-04-05
Faulty evidence error handling

setrofim:
Ensure correct handing of various errors in the provided evidence during verification. The general idea is that problems with the service setup should be communicted as errors (ultimately resulting in a 500 response from the REST frontend), while problems with the evidence should be communicated via an appropriately composed attestation result (with in a 200 response from the frontend).

This is done by creating an error type specifically to indicate bad evidnece, and handling it differently from others. Liminations in how errors can be represented in AR4SI are mitegated by providing additional information a a "problem" policy claim.

### DIFF
--- a/integration-tests/tests/test_end_to_end.tavern.yaml
+++ b/integration-tests/tests/test_end_to_end.tavern.yaml
@@ -65,7 +65,7 @@ stages:
           extra_kwargs:
             scheme: '{scheme}'
             evidence: '{evidence}'
-        - function: checkers:expected_result
+        - function: checkers:compare_to_expected_result
           extra_kwargs:
             expected: data/results/{scheme}.{expected}.json
             verifier_key: data/keys/verifier.jwk
@@ -105,7 +105,7 @@ stages:
     response:
       status_code: 200
       verify_response_with:
-        - function: checkers:expected_result
+        - function: checkers:compare_to_expected_result
           extra_kwargs:
             expected: __generated__/expected/{scheme}.{expected}.server-nonce.json
             verifier_key: data/keys/verifier.jwk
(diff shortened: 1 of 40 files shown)


setrofim:
This addresses https://github.com/veraison/services/issues/107 and  https://github.com/veraison/services/issues/109

integration-tests/tests/test_end_to_end.tavern.yaml
```diff
@@ -23,6 +23,9 @@ marks:
       vals:
         - [psa, p1, good, full, ec.p256, good]
         - [psa, p1, good, mini, ec.p256, good]
+        - [psa, p1, missingclaims, mini, ec.p256, noident]
```
thomas-fossati, line 26: Why is this mapped to 99?  Is the signature incorrect?

integration-tests/tests/test_end_to_end.tavern.yaml
```diff
@@ -23,6 +23,9 @@ marks:
...
+        - [psa, p1, good, mini, bad, badcrypto]
+        - [psa, p1, badinstance, full, ec.p256, badinstance]
```
thomas-fossati, line 28: ditto: why 99? If the identity is not known (i.e., the trust anchor lookup fails) shouldn't we return instance-identity=1 or 97?

integration-tests/tests/test_end_to_end.tavern.yaml
thomas-fossati: as per f2f discussion, this needs a new AR4SI code point.

thomas-fossati (approved):
LGTM

PR #146, opened 2023-05-02
Correct Parsectpm ClassID to be a UUID string

yogeshbdeshpande:


ionut-arm (approved):
🙏🏻 thanks for the fix!

thomas-fossati (approved):

thomas-fossati:
> @thomas-fossati I have addressed the point and added more clarity to the test case!

thanks!

PR #148, opened 2023-05-04
Add the common utility function for key extraction

### LINKED ISSUE #142: BUG: Factor out common public key extraction components from all schemes, using TA endorsements of PEM key
I believe we have used this in other places: should it be factored out?

_Originally posted by @thomas-fossati in https://github.com/veraison/services/pull/140#discussion_r1179266584_

yogeshbdeshpande:
Fixes #142

### DIFF
--- a/scheme/common/utils.go
+++ b/scheme/common/utils.go
@@ -39,8 +39,8 @@ func GetMandatoryPathSegment(key string, fields map[string]*structpb.Value) (str
 	return segment, nil
 }
 
-func GetPubKeyFromPEMEncodedPubKey(ta []byte) (crypto.PublicKey, error) {
-	block, rest := pem.Decode(ta)
+func DecodePemSubjectPubKeyInfo(key []byte) (crypto.PublicKey, error) {
+	block, rest := pem.Decode(key)
 	if block == nil {
 		return nil, errors.New("could not extract trust anchor PEM block")
 	}
--- a/scheme/common/utils.go
+++ b/scheme/common/utils.go
@@ -9,7 +9,6 @@ import (
 	"errors"
 	"fmt"
 
-	"github.com/veraison/services/log"
 	"google.golang.org/protobuf/types/known/structpb"
 )
 
@@ -43,7 +42,6 @@ func GetMandatoryPathSegment(key string, fields map[string]*structpb.Value) (str
 func GetPublicKeyFromTa(ta []byte) (crypto.PublicKey, error) {
 	block, rest := pem.Decode(ta)
 	if block == nil {
-		log.Error("Could not get TA PEM Block in GetPublicKeyFromTa")
 		return nil, errors.New("could not extract trust anchor PEM block")
 	}
 
--- a/scheme/common/utils.go
+++ b/scheme/common/utils.go
@@ -39,6 +39,7 @@ func GetMandatoryPathSegment(key string, fields map[string]*structpb.Value) (str
 	return segment, nil
 }
 
+// DecodePemSubjectPubKeyInfo decodes a PEM encoded SubjectPublicKeyInfo
 func DecodePemSubjectPubKeyInfo(key []byte) (crypto.PublicKey, error) {
 	block, rest := pem.Decode(key)
 	if block == nil {
--- a/scheme/common/utils.go
+++ b/scheme/common/utils.go
@@ -42,9 +42,8 @@ func GetMandatoryPathSegment(key string, fields map[string]*structpb.Value) (str
 
 func GetPublicKeyFromTa(ta []byte) (crypto.PublicKey, error) {
 	block, rest := pem.Decode(ta)
-
 	if block == nil {
-		log.Error("Could not get TA PEM Block ValidateEvidenceIntegrity")
+		log.Error("Could not get TA PEM Block in GetPublicKeyFromTa")
 		return nil, errors.New("could not extract trust anchor PEM block")
 	}
 
--- a/scheme/common/utils.go
+++ b/scheme/common/utils.go
@@ -3,9 +3,13 @@
 package common
 
 import (
+	"crypto"
+	"crypto/x509"
+	"encoding/pem"
 	"errors"
 	"fmt"
 
+	"github.com/veraison/services/log"
 	"google.golang.org/protobuf/types/known/structpb"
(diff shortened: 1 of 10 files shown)


scheme/common/utils.go
```diff
@@ -35,3 +39,26 @@ func GetMandatoryPathSegment(key string, fields map[string]*structpb.Value) (str
 
 	return segment, nil
 }
+
+func GetPublicKeyFromTa(ta []byte) (crypto.PublicKey, error) {
```
thomas-fossati, line 43: the function name is a bit misleading: since what the function does is extracting a public key from some PEM encoded data that is supposedly encoding a SubjectPublicKeyInfo.  A better name is, e.g., `DecodeSubjectPublicKeyInfo()`

yogeshbdeshpande: It is not just a subjectPublicKeyInfo, but a full PEM File.

I had it initially had it as:
`GetPublicKeyFromPemTa`

I am ok, to be more specific about it been a full pem ta with header as well

scheme/common/utils.go
```diff
@@ -35,3 +39,26 @@ func GetMandatoryPathSegment(key string, fields map[string]*structpb.Value) (str
...
+
+	if block.Type != "PUBLIC KEY" {
+		return nil, fmt.Errorf("unsupported key type: %q", block.Type)
+	}
+
+	pk, err := x509.ParsePKIXPublicKey(block.Bytes)
+	if err != nil {
+		return nil, fmt.Errorf("unable to parse public key: %w", err)
+	}
+	return pk, nil
```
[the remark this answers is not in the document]
yogeshbdeshpande: After checking x.509 error return parameters for ECDSA, I think it does add a value as otherwise the errors returned will be like below:
x509: invalid ecdsa parameter
x509: failed to unmarshal elliptic curve point

So with the added information it sets the top level context with the specific reason as to why parse failed.

scheme/common/utils.go
thomas-fossati: > It is not just a subjectPublicKeyInfo, but a full PEM File.

hmm `x509.ParsePKIXPublicKey` only accepts SPKI, so yes it's just correctly encoded SPKIs that this function will accept.

yogeshbdeshpande: No, what I meant was we also check for valid PEM Headers in this function, in addition to DecodeSubjectPublicKeyInfo
hence I mentioned the name as given above!

yogeshbdeshpande: Or may be GetPublicKeyFromPemPublic()

scheme/common/utils.go
```diff
@@ -35,3 +39,26 @@ func GetMandatoryPathSegment(key string, fields map[string]*structpb.Value) (str
 
 	return segment, nil
 }
+
+func GetPublicKeyFromTa(ta []byte) (crypto.PublicKey, error) {
+	block, rest := pem.Decode(ta)
+
+	if block == nil {
+		log.Error("Could not get TA PEM Block ValidateEvidenceIntegrity")
```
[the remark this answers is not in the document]
yogeshbdeshpande: Have done that I assume you meant spurious line on line 45?

scheme/common/utils.go
thomas-fossati: all this function is doing is providing a decoder for https://www.rfc-editor.org/rfc/rfc7468#section-13, so `DecodeTextualSPKI()` ?

scheme/common/utils.go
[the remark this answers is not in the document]
thomas-fossati: no, I meant to remove the log line altogether (log lines don't belong in library code)

scheme/common/utils.go
yogeshbdeshpande: I think: GetPublicKeyFromPemPublic() precisely conveys that meaning in a simple form rather than
a cryptic use of SPKI ?

https://en.wikipedia.org/wiki/Privacy-Enhanced_Mail

scheme/common/utils.go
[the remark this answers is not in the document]
yogeshbdeshpande: Ah, OK, I will remove that:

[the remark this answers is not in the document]
yogeshbdeshpande: Done!

scheme/common/utils.go
yogeshbdeshpande: It will also be not very clear in plugin:

I think, PEM format can convey a lot of things, like Certs etc.

if we refer to:
https://www.rfc-editor.org/rfc/rfc7468#section-4

It has a clear reference to PublicKey()

So 
GetPubKeyFromPemEncodePubKey()

would be more clear

yogeshbdeshpande: @thomas-fossati I have renamed  it. Please have a look!

yogeshbdeshpande: it is ok, I think

thomas-fossati: > it is ok, I think

we agree to disagree

yogeshbdeshpande:
@thomas-fossati Do you have any other comments apart from naming. I plan to submit the code as is, considering the correct label used as per the spec section 4: https://www.rfc-editor.org/rfc/rfc7468#section-4

scheme/common/utils.go
```diff
@@ -35,3 +38,24 @@ func GetMandatoryPathSegment(key string, fields map[string]*structpb.Value) (str
 
 	return segment, nil
 }
+
+func GetPubKeyFromPEMEncodedPubKey(ta []byte) (crypto.PublicKey, error) {
```
thomas-fossati, line 42: since this is shared code, we should add some suitable godoc comment.

also please change `ta` to something more meaningful meaningful

thomas-fossati:
> @thomas-fossati Do you have any other comments apart from naming. I plan to submit the code as is, considering the correct label used as per the spec section 4: https://www.rfc-editor.org/rfc/rfc7468#section-4

Not sure I understand what you mean.  Reading the labels defined in §4, it'd seem you are planning to use SubjectPublicKeyInfo?  If so, I suggest `PEMDecodeSubjectPublicKeyInfo()`

thomas-fossati (approved):

PR #151, opened 2023-05-12
deploy/docker: fix individual container make targets

setrofim:
Makefile targets for the individual containers (e.g. "vts" depend o resolution of ".built/%-container", which, in turn, depends on the resolution of ".built/%-image". The latter has $DEPLOY_DEST as its prerequisite, and, it appears, does not resolve if that prerequisite does not yet exist.

To fix this, add "deploy" (which creates $DEPLOY_DEST" as an explicit prerequisite for container and image targets, head of  the ".built/*" prerequisites.


thomas-fossati (approved):
thanks!

PR #152, opened 2023-05-12
Parsec TPM plugin: fix encoding of the `akpub` claim

thomas-fossati:
Needs to be Base64 URL-safe w/out padding

### DIFF
--- a/scheme/parsec-tpm/evidence_handler.go
+++ b/scheme/parsec-tpm/evidence_handler.go
@@ -277,7 +277,7 @@ func populateAttestationResult(
 	if err != nil {
 		return fmt.Errorf("unable to marshal public key: %w", err)
 	}
-	pubkey := base64.StdEncoding.EncodeToString(kd)
+	pubkey := base64.RawURLEncoding.EncodeToString(kd)
 	appraisal.AppraisalExtensions.VeraisonKeyAttestation = &map[string]interface{}{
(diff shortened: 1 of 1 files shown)


yogeshbdeshpande:
please create a github issue to link, to this change, explaining why this extension needs to be URL safe?
Are we planning to use this extension as a URI or for file searchin?

thomas-fossati:
> please create a github issue to link, to this change, explaining why this extension needs to be URL safe?

There's no need for an issue, we just lost [this bit](https://github.com/CCC-Attestation/attested-tls-poc/issues/16#issuecomment-1443600947) of the original discussion while implementing the Veraison plugin.  This results in impedance mismatch with the [c-ear implementation](https://github.com/veraison/c-ear/blob/3aa7c518a89df14955c8ef18b49ea8c1989d6f13/src/utils.c#L40) which, correctly, assumes `akpub` to be B64 Url-safe w/out padding.

This is just a quick fix to unblock the demo.
We need to come back here and fix it properly once we are done with https://github.com/veraison/ear/issues/30.

yogeshbdeshpande:
@thomas-fossati Thanks for setting up the context.

Proper fix to be doe by : https://github.com/veraison/services/issues/153

Changes LGTM as an interim fix!

yogeshbdeshpande (approved):

PR #155, opened 2023-05-17
update license scan tooling

thomas-fossati:
Since we simplified the modules structure since we last run the scan...


yogeshbdeshpande (approved):
LGTM

Issue #101, opened 2022-12-12
feature: Add default policy IDs to schemes.


setrofim:
Currently, the policy ID is used identify the third party-provided policy that is applied on top of the regular evaluation, if any. EAT Attestation Result draft, uses the term "policy" in a wider sense to encompass what Veraison calls a "scheme" as well. In order for our result to be better aligned with EAR, we should set the policy ID to identify the scheme as well.

See also comments here: https://github.com/thomas-fossati/draft-ear/pull/1

## Proposal

Change the Policy ID to be a URI with the scheme `policy` and the path indicating how the attestation result  was evaluated. The first segment of the path would indicate the Veraison scheme (plugin) used. Additional path segments would indicate the third-party policies that were used, the order of the segments matching the order in which these policies were applied (note: currently, Veraison allows only one third party policy per evaluation.)

### Examples

- `policy://tpm_enacttrust`: result generated using TPM_ENACTTRUST scheme, with no additional policy applied.
- `policy://psa_iot/mypolicy`: result generated using PSA_IOT scheme with additional policy identified as "mypolicy" in the Veraison policy store, applied.
- (hypothetical) `policy://tcg_dice/policy1/policy2`: result generated using TCG_DICE scheme with additional policies "policy1" and "policy2" applied (in that order, with "policy2" potentially overriding decisions made by "policy1").

setrofim:
Implemented by df713d4d

PR #168, opened 2023-06-23
Consistent AppraisalPolicyID and scheme-specific policies

setrofim:
"Policy" as defined by the RATS architecture[1], and, therefore, the ear.appraisal-policy-id entry n EAR[2], maps onto the combination of attestation scheme and policy in Veraison.

This means that, when a policy is not used, the ear.appraisal-policy-id field should be set to reflect the attestation scheme. If a policy is used, the field should be set to reflect both, the scheme and the policy.

Additionally, up to this point, the policy manager allowed only one active policy per tenant. Differentiation between schemes, if necessary, could be performed within the policy rules. This commit changes this so that the polices are now managed based on both, the tenant and the scheme. This means that policies for different schemes can be updated independently by the tenant.

[1]: https://datatracker.ietf.org/doc/html/draft-ietf-rats-architecture-05
[2]: https://www.rfc-editor.org/rfc/internet-drafts/draft-fv-rats-ear-00.html

This resolves: https://github.com/veraison/services/issues/101

Note: in that issue, the examples are incorrect in that they show the sheme to be encoded in the authority. This implementation follows the text (and the intent) of the proposal, and the scheme is encoded in the first path segement. Further, this  additionally specifies that the path is rootless.


policy/README.md
```diff
@@ -31,3 +31,92 @@ The following policy agent configuration directives are currently supported:
...
+For example:
+
+- `policy:PSA_IOT/0:base:v1/0:extra:v1`: the appraisal has been created using
+  "PSA_IOT" scheme, it was then updated by a policy with the individual policy
+  id `0:base:v1`, followed by a policy with the individual policy ID
+  `0:extra:v1`.
+
+### policy store key
+
+Policies are stored, retried from, and updated in the policy store using a key.
```
yogeshbdeshpande, line 84:
```suggestion
Policies are stored, retrieved from, and updated in the policy store using a key.
```

policy/policy.go
```diff
@@ -2,14 +2,90 @@
...
+	"strings"
+)
+
+// PolicyID identifies a spacific policy. This is used to retrieve the policy
+// from the store.
+type PolicyID struct {
+	// TenantId  is the ID of the tenant that owns this policy.
+	TenantId string `json:"tenant_id"`
+
+	// Scheme is the name of the scheme with witch this policy is associated
```
yogeshbdeshpande, line 18:
```suggestion
	// Scheme is the name of the scheme to which this policy is associated
```

policy/store.go
```diff
@@ -68,13 +68,13 @@ func (o *Store) Update(id, rules string) error {
...
 
-	return o.KVStore.Add(id, string(newPolicyBytes))
+	return o.KVStore.Add(id.StoreKey(), string(newPolicyBytes))
 }
 
 // Get returns the slice of all Policies associated with he specified ID. Each
 // Policy represents a different version of the same logical policy.
-func (o *Store) Get(id string) ([]Policy, error) {
-	vals, err := o.KVStore.Get(id)
+func (o *Store) Get(id PolicyID) ([]Policy, error) {
```
yogeshbdeshpande, line 76: Please change the typo on line 74

policy/store.go
```diff
@@ -152,8 +173,8 @@ func (o *Store) GetLatest(id string) (Policy, error) {
 }
 
 // Del removes all policy versisions associated with the specfied id.
-func (o *Store) Del(id string) error {
-	return o.KVStore.Del(id)
+func (o *Store) Del(id PolicyID) error {
```
yogeshbdeshpande, line 176: Please correct the typo on line 175

vts/appraisal/appraisal.go
```diff
@@ -5,27 +5,32 @@ package appraisal
...
 
-func New(tenantID string, nonce []byte, submodName string) *Appraisal {
+func New(tenantID string, nonce []byte, scheme string) *Appraisal {
 	appraisal := Appraisal{
+		Scheme: scheme,
 		EvidenceContext: &proto.EvidenceContext{
 			TenantId: tenantID,
 		},
-		Result: ear.NewAttestationResult(submodName, config.Version, config.Developer),
+		Result: ear.NewAttestationResult(scheme, config.Version, config.Developer),
```
yogeshbdeshpande, line 33: I have slight problem modifying this interface to EAR.

A scheme is replaced by submodName. 

There could be cases, when a single Scheme can appraise multiple SubMods and so we should retain submodName (as done previously) and add a specific Argument to EAR NewAttestationResult that lets one set the PolicyID for the subMod

yogeshbdeshpande (commented):
Left some comments.

vts/appraisal/appraisal.go
setrofim: The interface of `NewAttestationResult` isn't changing here. I'm renaming the argument that is being passed to it. That is not part of the interface. The corresponding parameter name is still `submodName`. 

The parateter of `Appraisal.New` is changing, however that is to more accurately reflect it's current useaged -- it is now used not only to instantiate an attestation result, but also initialize the policy ID (where it is not used as a submod name).

In the context of how `Appraisal.New` is used, it doesn't make sense to pass anything other than the scheme for this parameter, as is done early on, before eveidence handling has begun.

It is not the job of `Appraisal.New` to full initialize the submods of the attestation result. The only reason it needs a submod name at all is because an attestation result is required to have a least one submod, and we need to be able to return an attestation result early on.

The use case you're describing will need to be handled by the scheme-specific `EvidenceHandler` (probably inside `AppraiseEvidence`), where it would rename and/or add additional submods into the attestation result.

A case can be made for adding a new stage to our verification pipeline to be able to query the correct scheme-specific submod names, however that is a significant modification to the current implementation, involving the addition of a new API to the `IEvidenceHandler` interface, and necessitating corresponding changes in the existing plugin implemenations.  This is well outside the scope of the feature being implemented by this pull request.

yogeshbdeshpande (approved):
LGTM!

policy/policy.go
```diff
@@ -2,14 +2,90 @@
...
+			"bad policy store key %q: want 3 :-separated parts, found %d",
+			s, len(parts),
+		)
+	}
+
+	pid.TenantId = parts[0]
+	pid.Scheme = parts[1]
+	pid.Name = parts[2]
+
+	return pid, pid.Validate()
```
yogeshbdeshpande, line 42: From Line 12 to Line 65 should go to file id.go and the file id.go should be renamed to policyID.go

policy/policy.go
```diff
@@ -2,14 +2,90 @@
...
 	// Version gets bumped every time a new policy with existing ID is added to the store.
 	Version int32 `json:"version"`
 
 	// Rules of the policy to be interpreted and execute by the policy agent.
 	Rules string `json:"rules"`
 }
+
+// VersionedName returns a string identifier for the policy which is derived
+// from its policy ID and version (and is therefore unique across versions).
+func (o Policy) VersionedName() string {
```
yogeshbdeshpande, line 80: To be consistent, with the convention, why should a Policy Versioned Name not comprise of
PolicyID and Version, meaning why does the Versioned Name, not include the full PolicyID, which includes Scheme as well ?

setrofim: because `VersionedName` is used to identify the policy inside the `appraisal-policy-id`, which already contains the scheme.

yogeshbdeshpande: If we construct a consistent name, here which includes Tenant, scheme, name and version then we can overwrite the AppriasalPolicyID from the default scheme name at init to this name completely.

This way the VersionedName becomes consistent and does not rely on who is using this?
This is a public API afterall...

setrofim: we could, however, then:
- `appraisal-policy-id`  fomat will be incosinstent beteween cases when a policy when it was applied or it wasn't
- we increase redundancy in the `appraisal-policy-id` when we start supporting multiple policies.

yogeshbdeshpande: Yes, I thought a bit about it, if we state that the Versioned Policy ID (appraisal policy) is a combination of
Scheme, Tenant, name and version in that order then we can start with Scheme (default) and then replace with the actual `PolicyID` without losing the consistency ?

setrofim: True, however, this doesn't solve the second problem.

yogeshbdeshpande: I do not necessarily understand when you state increasing redundancy specifically. Also at the most I can understand per sub-mod specific AppraisalPolicyID which will entail  adding the subModName followed by the Scheme to provide the required uniqueness, in the PolicyID construct when setting the final `appraisal-policy-id`

setrofim: You would have duplication of scheme for each policy: `policy:SCHEME:0:base:v1/SCHEME:0:extra:v1`.

Conceptually: _either_ the policy generates the entire Appraisal ID (overriding the one set in cases where there is no policy), _or_ the policy only generates the part of the Appraisal ID  that represents itself.

If the policy generates the entire thing, then we have a problem in the mulitiple policy case, as they're unaware of each other, and the  last policy would simply override the id set py the previousl policies losing information.

If the policy only generates its part of the overall ID, then  the scheme should not be included in that part because the scheme is not part of the policy (even though a policy is tied to a  particular scheme, the scheme is _not_ tied to that policy).

setrofim: > per sub-mod specific AppraisalPolicyID

This is possible in our current pipleline -- submods are handled on per-scheme basis. The pipeline has no awareness of individual submods, and the policy agent operates on the attestation result as a whole, not individual submod appraisals.

yogeshbdeshpande: To my knowledge  at the moment the interface does not support the multiple policy handling (per submod) at the moment.

We have two options, we construct one unique `appraisal-policy-id` from the set of policies.
This can be effectively handled internally.

OR I assume we will have an array of  `appraisal-policy-id` strings and each UpdatePolicyID will append to that existing array. So I do not see the overwrite use case, in any way.

In either case, adding Scheme to the policy is always a wise and safe thing, even if there is a bit of duplication!

setrofim: > OR I assume we will have an array of appraisal-policy-id strings and each UpdatePolicyID will append to that existing array. So I do not see the overwrite use case, in any way.

This is effectively what is happening now (execept it's a single string that is being appended to, rather than array). However if each policy's "id" is being appended to it, tha "id" does not need to include the scheme because the scheme is already present (and is the same for all policies).

setrofim: >In either case, adding Scheme to the policy is always a wise and safe thing, even if there is a bit of duplication!

The schem is already included in the Appraisal ID.  there is no point in adding it again for each policy. It does not make things in any way safer.

policy/README.md
```diff
@@ -31,3 +31,92 @@ The following policy agent configuration directives are currently supported:
...
+## Policy Identification
+
+There are three different ways of identifying a policy:
+
+### appraisal policy ID
+
+This is an identifier used in the attestation result as defined by [EAR
+RFC](https://www.rfc-editor.org/rfc/internet-drafts/draft-fv-rats-ear-00.html#name-ear-appraisal-claims).
+Note that in this case, "policy" is used in the sense of the [RATS
+architecture](https://datatracker.ietf.org/doc/html/draft-ietf-rats-architecture-05)
```
thomas-fossati, line 44: * s/.../in the sense of "Appraisal Policy for Evidence" as per RATS architecture/
* for the link use https://www.rfc-editor.org/rfc/rfc9334

policy/README.md
```diff
@@ -31,3 +31,92 @@ The following policy agent configuration directives are currently supported:
...
+using which the appraisal has been generated. The first segment of the path is
+the name of the scheme used to create the appraisal. The second segment, if
+present, is the individual policy ID (see below) of the policy that has been
+applied to the appraisal created by the scheme.
+
+For example:
+
+- `policy:TPM_ENACTTRUST`: the appraisal has been created using "TPM_ENACTTRUST" scheme, with
+  no additional policy applied.
+- `policy:PSA_IOT/0:opa:v1`: the appraisal has been created using "PSA_IOT"
```
thomas-fossati, line 59: For cascading policies the tenant ID is repeated in every segment, so it's worth factoring out into the first segment.  Especially considering the fact that tenant IDs may not be that short (e.g., UUIDs or other kinds of long-ish identifiers).

So something like: `policy:0:PSA_IOT/opa:v1`

thomas-fossati (commented):
thanks! a few comments inlined

policy/README.md
setrofim: This is now obsolite -- the individual policy ID (here in the format `[tenant]:[name]:[version]` will be replaced with the policy instances' UUID instead (as be discussion last Thursday).

setrofim:
This is now obsolite. Updated changes to policy id will be submitted as a part of larger policy management pull.

PR #170, opened 2023-06-28
policy: fix scheme name inside policy evaluation

setrofim:
The scheme name inside the OPA policy package was being set based on the attestation-format, which is no longer used. So, in practice, it was not being set correctly at all.

This commit:

- renames the "format" field to "scheme" to be consistent with the general terminology.
- populates this field based in scheme name passed onto input via a new parameter to the policy manager.

The upshot is that the scheme name is now accessible from inside the policy.


policy/README.opa.md
```diff
@@ -47,7 +47,7 @@ scheme-specific.
 `result` is a JSON object representing `proto.AttestationResult` that was
 generated by the scheme.
 
-`format` is the name of `proto.AttestationFormat` associated with the scheme.
+`scheme` is the name of the attestation scheme.
```
yogeshbdeshpande, line 50: All looks good here,  while you are making this change, shall we be more precise and change Line 40: to say:
`evidence` contains the JSON representation of `proto.EvidenceContext.Evidence`

yogeshbdeshpande (approved):
LGTM!

PR #180, opened 2023-07-27
Cleanup of vts.proto interface

### LINKED ISSUE #176: Tidy up vts proto interface to remove structure definitions
As part of moving Provisioning decoder plugins under `vts` the movement left few proto interface structures, which needs clean up.

This is taken as a separate task, as the movement has a wider impact on all plugin interfaces.

yogeshbdeshpande:
Remove unwanted proto messages from the vts.proto interface.

Fixes #176


setrofim (approved):
LGTM

PR #186, opened 2023-08-21
Implement auth

setrofim:
Implement API authentication and authorization.

- Define IAuthorizer interface that can be used to obtain a gin
  middleware handler that performs authorisation.
- Add a mechanism to obtain an IAuthorizer for a particular role base on
  Veraison configuration.
- Implement "passthrough" authorizer that duplicates existing behavior
  (no auth).
- Implement "basic" authorizer that does not rely on an external
  authorization server.
- Implement "keycloak" authorizer that uses Keycloak service for
  authentication.
- Update provisioning service to authorize based on "provisioner" role.
- Update management service to authorize based on "manager" role.
- Add the previously missing README for the management service.

**note**: Once the corresponding client-side support for auth has been added, there will be additional changes to update the `colci`/ `pocli` inside the docker deployment.


deployments/docker/veraison
```diff
@@ -9,20 +9,24 @@ function status() {
     local prov=$(_get_container_state provisioning-service)
     local verif=$(_get_container_state verification-service)
     local manage=$(_get_container_state management-service)
+    local keycloak=$(_get_container_state keycloak-service)
```
yogeshbdeshpande, line 12: Presumably: This has an impact on the diagram listed here:?
https://github.com/veraison/services/tree/main/deployments/docker

auth/README.md
```diff
@@ -0,0 +1,131 @@
...
+		log.Fatalf("Could not parse config: %v", err)
+	}
+
+    // Create new authorizer based on the loaded config.
+	authorizer, err := auth.NewAuthorizer(subs["auth"], log.Named("auth"))
+	if err != nil {
+		log.Fatalf("could not init authorizer: %v", err)
+	}
+
+    // Ensure the authorizer is terminated properly on exist
```
yogeshbdeshpande, line 107:
```suggestion
    // Ensure the authorizer is terminated properly on exit
```

auth/README.md
```diff
@@ -0,0 +1,131 @@
+## Overview
+
+This implements authentication and authorization for Veraison API.
+Authentication can be performed using the Basic HTTP scheme (with the `basic`
+backend), or using a Bearer token (with the `keycloak` backend). Once an API
+user is authenticated, authorization is role-based. See documentation for
```
yogeshbdeshpande, line 6:
```suggestion
user is authenticated, authorization is role-based. See documentation [here](link to documentation) for
```

auth/basic.go
```diff
@@ -0,0 +1,149 @@
...
+	if rawUsers := v.GetStringMap("users"); rawUsers != nil {
+		for name, rawUser := range rawUsers {
+			switch t := rawUser.(type) {
+			case map[string]interface{}:
+				newUser, err := newBasicAuthUser(t)
+				if err != nil {
+					return fmt.Errorf("invalid user %q: %w", name, err)
+
+				}
+				o.logger.Debugw("registred user",
```
yogeshbdeshpande, line 80:
```suggestion
				o.logger.Debugw("registered user",
```

deployments/docker/src/keycloak.conf
```diff
@@ -0,0 +1,45 @@
+# Basic settings for running in production. Change accordingly before deploying the server.
```
yogeshbdeshpande, line 1: May be worth adding a statement as to why these configurations are commented and not used in the conf file.

deployments/docker/src/keycloak.docker
```diff
@@ -0,0 +1,17 @@
+FROM quay.io/keycloak/keycloak:22.0.1 as keycloak-builder
+
+WORKDIR /opt/keycloak
+# note: for development set up early; use proper certification in production.
+RUN keytool -genkeypair -storepass password -storetype PKCS12 -keyalg RSA -keysize 2048 \
+	-dname "CN=server" -alias server -ext "SAN:c=DNS:localhost,IP:127.0.0.1" \
```
yogeshbdeshpande, line 6: should the server on  line 6 be hard coded  or read from yaml file?

deployments/docker/src/load-config.mk
```diff
@@ -11,6 +11,7 @@ VTS_PORT ?= 50051
 PROVISIONING_PORT ?= 8888
 VERIFICATION_PORT ?= 8080
 MANAGEMENT_PORT ?= 8088
+KEYCLOAK_PORT ?= 11111
```
yogeshbdeshpande, line 14: Just a thought, this could be 8880 ?

provisioning/cmd/provisioning-service/README.md
```diff
@@ -6,6 +6,11 @@ configuration:
 - `provisoning`: provisioning service configuration. See [below](#provisioning-service-configuration).
 - `vts` (optional): Veraison Trusted Services backend configuration. See [trustedservices config](/vts/trustedservices/README.md#Configuration).
 - `logging` (optional): Logging configuration. See [logging config](/vts/log/README.md#Configuration).
+- `auth` (optional): API authentication and authorization mechanism
```
yogeshbdeshpande, line 9: The spelling of provisioning on line 6 needs correction

provisioning/cmd/provisioning-service/main.go
```diff
@@ -73,8 +74,19 @@ func main() {
 	provisioner := provisioner.New(vtsClient)
 
 	log.Infow("initializing provisioning API service", "address", cfg.ListenAddr)
+	authorizer, err := auth.NewAuthorizer(subs["auth"], log.Named("auth"))
```
yogeshbdeshpande, line 77: Presumably if the Line 41 `auth` subs is missing the NewAuthroizer will default to `passthrough` ?

auth/README.md
setrofim: There isn't  really a link here -- different services have different documentation, and trying to maintain a comprehensive set of links here is going to do more harm than good, as it is pretty much guaranteed to get out of sync.

deployments/docker/src/keycloak.conf
setrofim: They're are used -- they're commented because the specified values are the defaults, i.e. they are used without needing to be set explicitly. Having them commented shows what the default is, and makes it easy to change the setting. This is a well-known convention for Unix conf files.

But actually, this is a stray file I forgot to delete, as we're using the `keycloak.conf.template` for the deployment, not this. Will remove.

deployments/docker/src/keycloak.docker
setrofim: nah, this isn't really used. This sets up keys used for the HTTPS cert, but we're enabling and using HTTP (specifically to avoid having to deal with certs in the test deployment). There isn't really a way to disable HTTPS (that I've seen), so the keys need to generated even if they're not used.

deployments/docker/src/load-config.mk
setrofim: I specifically chose something different, since this isn't really part of Veraison -- this is an external service Veraison depends on (besides, its already difficult enough to remember the correct port to use for DEBUG_PORT, as they're all just 8's and 0's in various permutations).

provisioning/cmd/provisioning-service/main.go
setrofim: Yup: https://github.com/veraison/services/blob/setrofim-auth/auth/authorizer.go#L21

auth/README.md
```diff
@@ -0,0 +1,133 @@
...
+- `users`: this is a mapping of user names onto their bcrypt password hashes
+  and roles. The key of the mapping is the user name, the value is a futher
+  mapping for the details with the following fields:
+
+    - `password`: the bcrypt hash of the user's password.
+    - `roles`: either a single role or a list of roles associated with the
+      user. API authrization will be performed based on the user's roles.
+
+bcrypt hashes can be generated on the command line using `mkpasswd` utility,
+e.g.:
```
thomas-fossati, line 44: On Linux, bcrypt hashes can be generated on the command line using `mkpasswd`
utility, e.g.:

deployments/docker/src/veraison-realm.json
thomas-fossati, line 1: lord almighty, from whence it comes this unholy beast? 😄 😄 😄

thomas-fossati (approved):
LGTM, thanks!

auth/basic.go
```diff
@@ -0,0 +1,149 @@
...
+}
+
+func (o *BasicAuthorizer) GetGinHandler(role string) gin.HandlerFunc {
+	return func(c *gin.Context) {
+		o.logger.Debugw("auth basic", "path", c.Request.URL.Path)
+
+		userName, password, hasAuth := c.Request.BasicAuth()
+		if !hasAuth {
+			c.Writer.Header().Set("WWW-Authenticate", "Basic realm=veraison")
+			c.AbortWithStatus(http.StatusUnauthorized)
```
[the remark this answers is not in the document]
setrofim: Good point. Will add that.

Incidentally, we really need to think about how we're organising our code. There is enough commonality between the different REST API (verification, provisioning, etc) implementations, such as the `ReportProblem`, that it would make sense to have a common place for the code, rather than duplicationg it for each API.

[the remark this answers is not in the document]
thomas-fossati: > Good point. Will add that.

awesome

> Incidentally, we really need to think about how we're organising our code. There is enough commonality between the different REST API (verification, provisioning, etc) implementations, such as the `ReportProblem`, that it would make sense to have a common place for the code, rather than duplicationg it for each API.

➕1

auth/keycloak.go
```diff
@@ -0,0 +1,109 @@
...
+		Realm: "veraison",
+	}
+
+	loader := config.NewLoader(&cfg)
+	if err := loader.LoadFromViper(v); err != nil {
+		return err
+	}
+
+	o.config = ginkeycloak.KeycloakConfig{
+		Url:                fmt.Sprintf("http://%s:%s", cfg.Host, cfg.Port),
```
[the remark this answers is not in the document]
thomas-fossati: > (note to self: we need an issue for moving all the service endpoints to HTTPS.)

Added https://github.com/veraison/services/issues/191

auth/keycloak.go
```diff
@@ -0,0 +1,109 @@
...
+		return roleOK
+	}
+}
+
+func mapTenantID(jsonWebToken *jwt.JSONWebToken, keyCloakToken *ginkeycloak.KeyCloakToken) error {
+	var claims map[string]interface{}
+
+	// note: this mapper function will only be called once the JWT had
+	// alreadybeen verified by ginkeycloak, so extracting claims without
+	// verification here is, in fact, safe.
```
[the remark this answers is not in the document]
setrofim: haven't tried that, as the only way to easlily get a JWT than can be used for this is from keycloak it self. We can test it out at some point, but I would argue that is a security issue with keycloak.

I.e. it is somehting that needs to be tested and secured for a particular deployment, not withithin veraison code.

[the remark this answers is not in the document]
thomas-fossati: > [...] but I would argue that is a security issue with keycloak.

yes


BTW, timely (but yet to be answered) question: https://github.com/keycloak/keycloak/discussions/22744

auth/basic.go
```diff
@@ -0,0 +1,154 @@
...
+					break
+				}
+			}
+		}
+
+		if gotRole {
+			log.Debugw("user authenticated", "user", userName, "role", role)
+		} else {
+			c.Writer.Header().Set("WWW-Authenticate", "Basic realm=veraison")
+			c.AbortWithStatus(http.StatusUnauthorized)
```
thomas-fossati, line 149: L149 should go (it's done in ReportProblem)

auth/keycloak.go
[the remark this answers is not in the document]
setrofim: oh, yeah. Thanks for finding that. Upvoted and subscribed.

auth/basic.go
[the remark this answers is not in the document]
setrofim: Raised https://github.com/veraison/services/issues/192

PR #188, opened 2023-08-31
scheme/tpm-enacttrust: fix expected trust anchor

setrofim:
Previously, the scheme expected the trust anchor to be base64-encoded PKIX Subject Public Key. This was in line with the early CORIM expectations.

Current CORIM spec defines a number of possible trust anchor types, none of which match that. This commit changes the expected key to be the PEM-encoded Subject Public Key instead (essentially, the same thing as before, but wrapped with PEM header/footer).


yogeshbdeshpande (approved):
LGTM!

PR #189, opened 2023-09-01
scheme/tpm-enacttrust: fix failed attestation result

setrofim:
Evidence processing is checking that the hash in the evidence matches the provisioned reference value, and sets status to affirming if it does.

However, prior to this, it was not doing anything otherwise. This means that in case of failed attestation, the result status remained as "none".

This commit sets status to contraindicated (and executables trust vector claim to unknown runtime) if reference value comparison fails.


yogeshbdeshpande (approved):
LGTM!

PR #197, opened 2023-09-15
Enchanced TA validation + CLI auth in docker deployment

setrofim:
- Update corim package dependency to latest, which feature updated trust
anchor representations that allow for tighter validation based on a
set of predefined types.
- In addition to the validation we get "for free" from the above, add
checks to endorsement decoders to make sure that the TA's are of the
expected type for the scheme.
- Upgrade the CLI tools in the docker deployment to the latest versions
that have auth support. This fixes end-to-end script which did not work
since the deployment had auth enabled.


deployments/docker/src/builder.docker
```diff
@@ -57,7 +57,7 @@ RUN go mod download &&\
     go install google.golang.org/protobuf/cmd/protoc-gen-go@v1.26 &&\
     go install google.golang.org/grpc/cmd/protoc-gen-go-grpc@v1.1 &&\
     go install github.com/mitchellh/protoc-gen-go-json@v1.1.0 &&\
-    go install github.com/veraison/corim/cocli@latest &&\
+    go install github.com/veraison/corim/cocli@eeb7bd48 &&\
```
yogeshbdeshpande, line 60: I thought we strongly enforce revisions now?

end-to-end/input/corim-src/comid-cca-refval.json
```diff
@@ -0,0 +1,109 @@
...
+              "type": "psa.refval-id",
+              "value": {
+                "label": "M3",
+                "version": "1.0.0",
+                "signer-id": "BwYFBAMCAQAPDg0MCwoJCBcWFRQTEhEQHx4dHBsaGRg="
+              }
+            },
+            "value": {
+              "digests": [
+                "sha-256:EwYFBAMCAQAPDg0MCwoJCBcWFRQTEhEQHx4dHBsaGRg="
```
yogeshbdeshpande, line 88: I am little confused:
We are still using `:` for digests here which (I thought ) is no longer recommended ?

See [comid-enacttrust-refval.json](https://github.com/veraison/services/pull/197/files#diff-807db7e8f06f7ad6e6e87db06f885be431ded69936cc7f8eb68abcc012a99e72) where you use `;`

scheme/psa-iot/test/ComidPsaRefValMultDigest.json
```diff
@@ -21,7 +21,7 @@
         "environment": {
           "class": {
             "id": {
-              "type": "PSA_IOT.impl-id",
+              "type": "psa.impl-id",
```
yogeshbdeshpande, line 24: is there a specific reason for this change?

This kind of breaks the nice alignment of the Store `key`-`value` symmetry?

Please reconsider!

deployments/docker/src/builder.docker
setrofim: That hash is assocated with a v2.0.0rc0 tag, however because it's a v2 tag for a modules without a `/v2`, using it won't work. (conversely, moving to `/v2` without publishing a release won't work either), it's a catch22 in the go module versioning system, it seems. So  have to use the hash here.

end-to-end/input/corim-src/comid-cca-refval.json
setrofim: Well, ether works, but yeah, technically should be a `;`. Will fix.

scheme/psa-iot/test/ComidPsaRefValMultDigest.json
setrofim: This change is required by `corim` (I think I mentioned this in our discussions earlier?). For whatever reason, `corim` templates explect `psa`. This is hard-coded: 

https://github.com/veraison/corim/blob/main/comid/classid.go#L168

I agree that we should align how we represent things in the stores vs the templeas (and better yet, we should get rid of these entirely unnecessary prefixes). However that would require making further changes to `corim` that orthogonal to what is being implemented here.

yogeshbdeshpande (approved):
LGTM!

PR #198, opened 2023-09-25
Nonce exposed to policy + fixes

setrofim:
Changes to allow policies access to the verification session nonce + fixes to nonce handling.

### DIFF
--- a/integration-tests/tests/test_cca_verify_challenge.tavern.yaml
+++ b/integration-tests/tests/test_cca_verify_challenge.tavern.yaml
@@ -1,0 +1,99 @@
+test_name: cca-verify-challenge
+
+marks:
+  - parametrize:
+      key:
+        #  Attestation scheme -- this is used to indicate how test cases should
+        #  be constructed (e.g. how the evidence token will be compiled.
+        - scheme
+        # Some attestation schemes (currently, only PSA) may support multiple
+        # profiles. If a scheme does not support multiple profiles, specify it
+        # as '_'.
+        - profile
+        # Which evidence description will be used to construct the evidence token.
+        - evidence
+        # The name of the endorsements spec within common.yaml
+        - endorsements
+        # Signing keys that will be used to construct the evidence. How this is
+        # used is dependent on the scheme.
+        - signing
+        # Expected structure of the returned EAR (EAT (Entity Attestation
+        # Token) Attestation Result).
+        - expected
+      vals:
+        - [cca, _, good, full, ccakeys, verify-challenge]
+
+includes:
+  - !include common.yaml
+
+stages:
+  - name: submit post request to the provisioning service successfully
+    request:
+      method: POST
+      url: http://{provisioning-service}/endorsement-provisioning/v1/submit
+      headers:
+        content-type: '{endorsements-content-type}' # set via hook
+        authorization: '{authorization}' # set via hook
+      file_body: __generated__/endorsements/corim-{scheme}-{endorsements}.cbor
+    response:
+      status_code: 200
+
+  - name: submit cca policy
+    request:
+      method: POST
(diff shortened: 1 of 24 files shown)


yogeshbdeshpande:
> Changes to allow policies access to the verification session nonce + fixes to nonce handling.

On the commit message:
Could you please clarify a bit more clearly about the change? Why this is needed and how this is helpful?

setrofim:
I believe the commit messages explain that. Please could you clarify what
needs to be added to them?

yogeshbdeshpande:
> fixes to nonce handling.

Two things:

I am not clear what is been stated: Why would the policy needs the session nonce ? 
What is the context? Is there a github issue that describes the context?


Also  fixes to nonce handling: 
is it the size correction or anything, I am not clear reading what is been fixed?

thomas-fossati (approved):
LGTM, thanks!

policy/agent.go
```diff
@@ -68,6 +68,7 @@ func (o *Agent) GetBackendName() string {
 // overwrite the result status or any of the values in the result trust vector.
 func (o *Agent) Evaluate(
 	ctx context.Context,
+	appraisalContext map[string]interface{},
```
yogeshbdeshpande, line 71: This change also means that the comment on Evaluate at line 66 needs to change!

yogeshbdeshpande, line 71: Could there be a better name as the name collides with AppraisalContext on the proto interface which means completely different data structure!

integration-tests/tests/test_cca_verify_challenge.tavern.yaml
```diff
@@ -0,0 +1,99 @@
+test_name: cca-verify-challenge
```
yogeshbdeshpande, line 1: Does this test checks negative case also, meaning Supplied `nonce` is not part of the token ?

yogeshbdeshpande (commented):
Reviewed

integration-tests/tests/test_cca_verify_challenge.tavern.yaml
setrofim: No, because that would be checked in the policy that is not part of the code being tested, but rather the test. What is being tested here is that the nonce associated with the session is now available to the policy (this also serves as an example and an answers one of the questions you asked above -- why is it useful of the nonce to be available to the policy), not the fact that the nonce in the token is being validated (it isn't outside this test).

policy/agent.go
setrofim: I don't believe so? The comment is no less valid than before. This new parameter doesn't change what the function is doing, or how it is doing it. This new parameter doesn't bear mentioning any more than the existing `ctx`, `submod`, `appraisal`, etc.

setrofim:
> I am not clear what is been stated: Why would the policy needs the session nonce ? What is the context? Is there a github issue that describes the context?

The policy should be able to make any decision as a scheme handler or the relying party verifier, as it is intened to be a potential per-deployment override/subsititute for either.  The fact that it did not already have access to it  is an omission (basically due to the fact that policies get called for each individual submod and  so don't have access to the outer attestation result, where the nonce happens to be. This is a bug being fixed, not a feature being added. There is no issue for this. I've added an additional paragraph to the second commit's message to clarify this. Let me know if something else being added.


> Also fixes to nonce handling: is it the size correction or anything, I am not clear reading what is been fixed?

- The fact the nonce was being lost form the attestationresult (see the first commit's message)
- The nonce size limits (indirectly by ugrading the `ear` depency --  see 3rd bullet in the second commit's message).

SimonFrost-Arm (approved):
LGTM

yogeshbdeshpande:
@setrofim Thank you for all the explainations and clarifications. LGTM now!

yogeshbdeshpande (approved):

PR #199, opened 2023-09-26
scheme: freshness check for PSA_IOT and CCA_SSD_PLATFORM

setrofim:
Update PSA_IOT and CCA_SSD_PLATFORM schemes to perform a freshness check as part of token integrity validation.

Freshness check is integral to attestation validation to prevent replay attacks. Thus far, we have not performed it as part of our verification pipeline, and instead left it the RP. However, PSA and CCA both mandate freshness claims, and so there is no reason why this check cannot be performed as part of the scheme evidence handling, therefore reducing the risk of the RP neglecting to consider freshness and taking an affirming attestation result from Veraison at face value.

### DIFF
--- a/integration-tests/data/results/cca.freshness-fail.json
+++ b/integration-tests/data/results/cca.freshness-fail.json
@@ -1,0 +1,18 @@
+{
+        "ear.appraisal-policy-id": "policy:CCA_SSD_PLATFORM",
+        "ear.status": "contraindicated",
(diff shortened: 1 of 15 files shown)


thomas-fossati (approved):
awesome, and much needed.  thanks!

yogeshbdeshpande (approved):
LGTM!

PR #200, opened 2023-09-27
verification: nonce size validation on session create

setrofim:
Up to this point, on session creation, nonceSize parameter was required to be a non-zero uint8 (i.e. in range 1..255), and the nonce parameter was only verified to be  a valid base64 string, with no size limit.

The nonce eventually ends up inside EAR's eat_nonce claim, and therefore must comport with the restriction associated with that claim, which is between 8 and 64 bytes long. This will result in a failed attestation result generation in the subsequent submit request.

This commit modifies nonceSize limit check to be aligned with eat_nonce definition, and adds an analogous check for the nonce parameter. This will insure that attempting to define a nonce of an invalid size during session creation will result in a failure to create the session.

### DIFF
--- a/verification/api/handler.go
+++ b/verification/api/handler.go
@@ -136,8 +136,8 @@ func parseNonceRequest(nonceParam, nonceSizeParam string) ([]byte, error) {
 	// a nonceSize was supplied, try to use it to mint a new nonce
 	if nonceSizeParam != "" {
 		nonceSize, err := aToU8(nonceSizeParam)
-		if err != nil {
-			return nil, errors.New("nonceSize must be in range 1..256")
+		if err != nil || nonceSize < 8 || nonceSize > 64 {
+			return nil, errors.New("nonceSize must be in range 8..64")
 		}
 		return mintNonce(nonceSize)
 	}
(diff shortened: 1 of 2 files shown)


verification/api/handler.go
```diff
@@ -148,6 +148,14 @@ func parseNonceRequest(nonceParam, nonceSizeParam string) ([]byte, error) {
 		return nil, errors.New("nonce must be valid base64")
 	}
 
+	nonceLen := len(nonce)
+	if nonceLen < 8 || nonceLen > 64 {
+		return nil, fmt.Errorf(
+			"nonce must be between 8 and 64 bytes long; got %d",
+			nonceLen,
```
yogeshbdeshpande, line 155:
```suggestion
			nonceLen
```

setrofim: That comma is required, as the closing paren is on a new line.

yogeshbdeshpande (approved):
LGTM!

verification/api/handler.go
yogeshbdeshpande: Do we need a newline though?

setrofim: "need" is a relative term, but yes. It breaks up a long line. And this makes it easier to modify the call later on (and results in neater diffs) if, e.g.,  we ant to add an additional parameter, or switch their order in the future.

yogeshbdeshpande: Sure, I am fine with this!

PR #201, opened 2023-09-27
scheme/cca-ssd-platform: zero-pad session nonce

setrofim:
Ensure that if the session nonce is less than 64 bytes, it gets zero-padded to that size, as this will be done to cca-realm-challenge inside the attestation token.

See https://developer.arm.com/documentation/den0137/latest, p 92

Thanks to @thomas-fossati for spotting this.


thomas-fossati (approved):
👍

PR #202, opened 2023-10-05
Scheme correction

yogeshbdeshpande:
Rename the scheme `tcg-dice` to its correct name `riot` as this is a a specific variant of DICE known as `riot`


setrofim (approved):
LGTM. Thanks!

PR #205, opened 2023-12-09
Modified Plugin Interface to handle Composite Attester Use Case

### LINKED ISSUE #206: Enhance Interface for Verification to handle Multiple Trust Anchor and Reference IDs
The current IEvidence Interface does not allow handling Multiple Trust Anchor IDs and 
Reference Value ID's, between VTS Base code and the Scheme Plugins.

An Attestation Scheme for Composite Attester, may need to handle multiple Trust Anchor ID's and Reference ID's
This issue tracks this enhancement.

yogeshbdeshpande:
Enhance Verification Interface (IEvidenceHandler) to handle multiple Reference Value and Trust Anchor IDs

Fixes #206


scheme/cca-ssd-platform/evidence_handler.go
```diff
@@ -46,13 +46,17 @@ func (s EvidenceHandler) SynthKeysFromTrustAnchor(tenantID string, ta *handler.E
 	return arm.SynthKeysFromTrustAnchors(SchemeName, tenantID, ta)
 }
 
-func (s EvidenceHandler) GetTrustAnchorID(token *proto.AttestationToken) (string, error) {
-	return arm.GetTrustAnchorID(SchemeName, token)
+func (s EvidenceHandler) GetTrustAnchorIDs(token *proto.AttestationToken) ([]string, error) {
+	ta, err := arm.GetTrustAnchorID(SchemeName, token)
+	if err != nil {
+		return nil, err
```
thomas-fossati, line 52: not sure this has any impact (we'd need to check all the calling sites) but there is a slight inconsistency here: we return `<nil>` when in other similar cases (e.g., see `handler/evidence_rpc.go`) we return `[]string{""}` instead.

vts/trustedservices/trustedservices_grpc.go
```diff
@@ -360,38 +360,44 @@ func (o *GRPC) GetAttestation(
...
-		"software-id", appraisal.EvidenceContext.ReferenceId,
-		"trust-anchor-id", appraisal.EvidenceContext.TrustAnchorId)
+		"software-id", appraisal.EvidenceContext.ReferenceIds,
+		"trust-anchor-id", appraisal.EvidenceContext.TrustAnchorIds)
 
-	endorsements, err := o.EnStore.Get(appraisal.EvidenceContext.ReferenceId)
-	if err != nil && !errors.Is(err, kvstore.ErrKeyNotFound) {
-		return o.finalize(appraisal, err)
-	}
+	var tendorsements []string
```
thomas-fossati, line 369: curiosity: what is the `t` in `tendorsements` exactly supposed to mean? 😄

vts/trustedservices/trustedservices_grpc.go
```diff
@@ -360,38 +360,44 @@ func (o *GRPC) GetAttestation(
...
+
+		endorsements, err := o.EnStore.Get(reference)
+		if err != nil && !errors.Is(err, kvstore.ErrKeyNotFound) {
+			return o.finalize(appraisal, err)
+		}
+
+		if len(endorsements) > 0 {
+			o.logger.Debugw("obtained endorsements", "endorsements", endorsements)
+		}
+		tendorsements = append(tendorsements, endorsements...)
```
thomas-fossati, line 380: Just a thought.  Why make it conditional?  It'd be also interesting to intercept the complementary condition, i.e., when an empty endorsement is returned, and maybe flag the anomaly.  In fact, I'm not sure we allow the returned ref-values to be empty...

vts/trustedservices/trustedservices_grpc.go
```diff
@@ -360,38 +360,44 @@ func (o *GRPC) GetAttestation(
...
-		"trust-anchor-id", appraisal.EvidenceContext.TrustAnchorId)
+		"software-id", appraisal.EvidenceContext.ReferenceIds,
+		"trust-anchor-id", appraisal.EvidenceContext.TrustAnchorIds)
 
-	endorsements, err := o.EnStore.Get(appraisal.EvidenceContext.ReferenceId)
-	if err != nil && !errors.Is(err, kvstore.ErrKeyNotFound) {
-		return o.finalize(appraisal, err)
-	}
+	var tendorsements []string
+	for _, reference := range appraisal.EvidenceContext.ReferenceIds {
```
thomas-fossati, line 370: Maybe, since it's an actual identifier, instead of `reference` call this `refvalID`

vts/trustedservices/trustedservices_grpc.go
```diff
@@ -418,17 +424,22 @@ func (c *GRPC) initEvidenceContext(
...
 
-func (c *GRPC) getTrustAnchor(id string) (string, error) {
-	values, err := c.TaStore.Get(id)
-	if err != nil {
-		return "", err
-	}
+func (c *GRPC) getTrustAnchors(id []string) ([]string, error) {
+
+	ta_values := make([]string, 0)
+	for _, taId := range id {
```
thomas-fossati, line 430: we use `taID` everywhere else in the codebase

```suggestion
	for _, taID := range id {
```

vts/trustedservices/trustedservices_grpc.go
```diff
@@ -418,17 +424,22 @@ func (c *GRPC) initEvidenceContext(
...
 }
 
-func (c *GRPC) getTrustAnchor(id string) (string, error) {
-	values, err := c.TaStore.Get(id)
-	if err != nil {
-		return "", err
-	}
+func (c *GRPC) getTrustAnchors(id []string) ([]string, error) {
+
+	ta_values := make([]string, 0)
```
thomas-fossati, line 429: golang wants us to use camelCase:

```suggestion
	taValues := make([]string, 0)
```

thomas-fossati, line 429: besides, you can just (more idiomatically) do:

```golang
    var taValues []string
```

vts/trustedservices/trustedservices_grpc.go
```diff
@@ -418,17 +424,22 @@ func (c *GRPC) initEvidenceContext(
...
+		values, err := c.TaStore.Get(taId)
+		if err != nil {
+			return []string{""}, err
+		}
 
-	if len(values) != 1 {
-		return "", fmt.Errorf("found %d trust anchors, want 1", len(values))
+		if len(values) != 1 {
+			return []string{""}, fmt.Errorf("found %d trust anchors, want 1", len(values))
+		}
```
thomas-fossati, line 438: This is an interesting restriction.  Do we want to remove it?  Or leave it in but document why we do it?

vts/trustedservices/trustedservices_grpc.go
yogeshbdeshpande: I meant totalEndorsements via `tendorsements` but may be it is not very helpful! 
I will rename it!

vts/trustedservices/trustedservices_grpc.go
yogeshbdeshpande: Returned Reference Values as Empty is a valid condition, it is up to the `scheme plugin` to flag it as an Error: if a specific implementation mandates a requirement  for a non-zero reference values!

yogeshbdeshpande: I will add a Debug to print the anomaly any way!

thomas-fossati: > Returned Reference Values as Empty is a valid condition, it is up to the `scheme plugin` to flag it as an Error: if a specific implementation mandates a requirement for a non-zero reference values!

hmm, I am not sure about this.  Before this change, an "empty" endorsement would be signalled by returning `kvstore.ErrKeyNotFound`

vts/trustedservices/trustedservices_grpc.go
thomas-fossati: ah, thanks!  Maybe we can use `endorsement` for each item and `endorsements` for the slice?

yogeshbdeshpande: in the inner block also, we return `endorsements` Hence I used slightly longer but `multEndorsements`

scheme/cca-ssd-platform/evidence_handler.go
yogeshbdeshpande: Agree, I have made it consistent! Thanks for pointing this out!

proto/evidence.proto
```diff
@@ -7,7 +7,7 @@ option go_package = "github.com/veraison/services/proto";
 
 message EvidenceContext {
   string tenant_id = 1 [json_name = "tenant-id"];
-  string trust_anchor_id = 2 [json_name = "trust-anchor-id"];
-  string reference_id = 3 [json_name = "reference-id"];
+  repeated string trust_anchor_ids = 2 [json_name = "trust-anchor-ids"];
+  repeated string reference_ids = 3 [json_name = "reference-ids"];
```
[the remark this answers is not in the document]
yogeshbdeshpande: Agree, as we are in relatively immature state, I have replaced, instead of deprecating the same!

handler/ievidencehandler.go
```diff
@@ -14,22 +14,22 @@ import (
 type IEvidenceHandler interface {
 	plugin.IPluggable
 
-	// GetTrustAnchorID returns a string ID used to retrieve a trust anchor
-	// for this token. The trust anchor may be necessary to validate the
-	// token and/or extract its claims (if it is encrypted).
-	GetTrustAnchorID(token *proto.AttestationToken) (string, error)
+	// GetTrustAnchorIDs returns an array of trust anchor strings(handles) used to retrieve a
+	// set of trust anchors for this token. The trust anchors may be necessary to validate the
+	// entire token and/or extract its claims (if it is encrypted).
```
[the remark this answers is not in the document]
yogeshbdeshpande: Have done the necessary change!

vts/trustedservices/trustedservices_grpc.go
yogeshbdeshpande: The existing `kvstore.ErrKeyNotFound` change still exists!

This should be the case, where the `key` exist in the Store but no associated `vals` detected for the specific `key`

vts/trustedservices/trustedservices_grpc.go
yogeshbdeshpande: agree ! i have changed it!

vts/trustedservices/trustedservices_grpc.go
yogeshbdeshpande: DONE!

vts/trustedservices/trustedservices_grpc.go
thomas-fossati: > This should be the case, where the `key` exist in the Store but no associated `vals` detected for the specific `key`

How can this happen?

yogeshbdeshpande: Checked the `kvstore` interface. Yes, this is not possible!

Hence removed the top level check for endorsements as this is superficial.

vts/trustedservices/trustedservices_grpc.go
yogeshbdeshpande: Added a comment to state the current restriction. We can revisit this if we want to support multiple TA's per 
TAID. Right now none of the schemes have this support!

yogeshbdeshpande: > Added a comment to state the current restriction. We can revisit this if we want to support multiple TA's per TAID. Right now none of the schemes have this requirement !

vts/trustedservices/trustedservices_grpc.go
```diff
@@ -360,38 +360,41 @@ func (o *GRPC) GetAttestation(
...
-		"trust-anchor-id", appraisal.EvidenceContext.TrustAnchorId)
+		"software-id", appraisal.EvidenceContext.ReferenceIds,
+		"trust-anchor-id", appraisal.EvidenceContext.TrustAnchorIds)
 
-	endorsements, err := o.EnStore.Get(appraisal.EvidenceContext.ReferenceId)
-	if err != nil && !errors.Is(err, kvstore.ErrKeyNotFound) {
-		return o.finalize(appraisal, err)
-	}
+	var multEndorsements []string
+	for _, refvalID := range appraisal.EvidenceContext.ReferenceIds {
```
setrofim, line 370: Note for the future (not necessary for this pull): Rather that looping over multiple requests to the store, we should add a Store API to handle multiple IDs  in a single request, e.g. so that his can just be `o.EnStore.GetMultiple(appraisal.EvidenceContext.ReferenceId)`.

vts/trustedservices/trustedservices_grpc.go
```diff
@@ -418,17 +421,23 @@ func (c *GRPC) initEvidenceContext(
...
-	values, err := c.TaStore.Get(id)
-	if err != nil {
-		return "", err
-	}
+func (c *GRPC) getTrustAnchors(id []string) ([]string, error) {
 
-	if len(values) != 1 {
-		return "", fmt.Errorf("found %d trust anchors, want 1", len(values))
+	var taValues []string //nolint
+	for _, taID := range id {
```
setrofim, line 427: Same note as above here -- we should add a `GetMultiple` to the store interface.

setrofim (approved):
LGTM

thomas-fossati (approved):
htanks for addressing my comments

vts/trustedservices/trustedservices_grpc.go
yogeshbdeshpande: @setrofim Valid point, I will raise an issue for the same!

yogeshbdeshpande: https://github.com/veraison/services/issues/207 raised to track this enhancement!

vts/trustedservices/trustedservices_grpc.go
yogeshbdeshpande: https://github.com/veraison/services/issues/207 raised to track this enhancement!

Issue #214, opened 2024-03-20
GetTrustAnchorID and ReferenceID Fetch should be symmetrical


yogeshbdeshpande:
In the current scheme of things, with PR #211 , GetTrustAnchorID is anchored on StoreHandler while ReferenceID fetch is anchored with EvidenceHandler.

This asymmetry needs to be addressed!

setrofim:
Addressed by https://github.com/veraison/services/pull/216

PR #216, opened 2024-04-24
handler: add GetRefValueIDs to IStoreHandler 

setrofim:
Move reference value ID extraction from evidence into store handler.
Prior to this, it was done inside evidence handler as part of claims
extraction.

This ensure that ID generation on both provisioning and verification
paths is handled in the same place, and is symmetrical with trust anchor
ID generation.

This also means that ExtractClaims method is now responsible _only_ for
claim extraction. ExtractedClaims structure is removed, and the method
now returns the map[string]interface{} claims set (ExtractedClaims
combined that with reference IDs).

This addresses: https://github.com/veraison/services/issues/214

### DIFF
--- a/handler/istorehandler.go
+++ b/handler/istorehandler.go
@@ -14,10 +14,19 @@ import (
 type IStoreHandler interface {
 	plugin.IPluggable
 
-	// GetTrustAnchorIDs returns an array of trust anchor identifiers used
+	// GetTrustAnchorIDs returns a slice of trust anchor identifiers used
 	// to retrieve the trust anchors associated with this token. The trust anchors may be necessary to validate the
 	// entire token and/or extract its claims (if it is encrypted).
 	GetTrustAnchorIDs(token *proto.AttestationToken) ([]string, error)
+
+	// GetRefValueIDs returns a slice of identifiers used to retrieve
+	// reference values for an attestation scheme, using the claims
+	// extracted from attestation token and the associated trust anchors.
+	GetRefValueIDs(
+		tenantID string,
+		trustAnchors []string,
+		claims map[string]interface{},
+	) ([]string, error)
 
 	// SynthKeysFromRefValue synthesizes lookup key(s) for the
 	// provided reference value endorsement.
--- a/scheme/cca-ssd-platform/evidence_handler.go
+++ b/scheme/cca-ssd-platform/evidence_handler.go
@@ -36,7 +36,7 @@ func (s EvidenceHandler) GetSupportedMediaTypes() []string {
 func (s EvidenceHandler) ExtractClaims(
 	token *proto.AttestationToken,
 	trustAnchors []string,
-) (*handler.ExtractedClaims, error) {
+) (map[string]interface{}, error) {
 
 	var ccaToken ccatoken.Evidence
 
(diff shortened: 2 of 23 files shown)


vts/trustedservices/trustedservices_grpc.go
```diff
@@ -376,13 +376,18 @@ func (o *GRPC) GetAttestation(
 		return o.finalize(appraisal, err)
 	}
 
-	appraisal.EvidenceContext.Evidence, err = structpb.NewStruct(extracted.ClaimsSet)
+	referenceIDs, err := stHandler.GetRefValueIDs(token.TenantId, tas, extracted)
+	if err != nil {
+		return o.finalize(appraisal, err)
+	}
+
+	appraisal.EvidenceContext.Evidence, err = structpb.NewStruct(extracted)
```
yogeshbdeshpande, line 384: Earlier we used to call line 371 variable as `extracted` as it was a composite structure.

is it not better if we call the variable as  `claimSet` and use the same at Line 384 appropriately ?

handler/istorehandler.go
```diff
@@ -14,11 +14,20 @@ import (
 type IStoreHandler interface {
 	plugin.IPluggable
 
-	// GetTrustAnchorIDs returns an array of trust anchor identifiers used
+	// GetTrustAnchorIDs returns a slice of trust anchor identifiers used
 	// to retrieve the trust anchors associated with this token. The trust anchors may be necessary to validate the
 	// entire token and/or extract its claims (if it is encrypted).
 	GetTrustAnchorIDs(token *proto.AttestationToken) ([]string, error)
 
+	// GetRefValueIDs returns a slice of refrence value identifiers used token
```
yogeshbdeshpande, line 22:
```suggestion
	// GetRefValueIDs returns a slice of reference value identifiers used token
```

yogeshbdeshpande, line 22:
```suggestion
	// GetRefValueIDs returns a slice of refrence value identifiers used 
```

yogeshbdeshpande, line 22: how about the following text:

// GetRefValueIDs returns the list of reference value identifiers used to retrieve the reference values
//  for an attestation scheme, using the claims extracted from attestation token

scheme/cca-ssd-platform/evidence_handler.go
```diff
@@ -58,18 +57,12 @@ func (s EvidenceHandler) ExtractClaims(
 			"could not convert realm claims: %w", err))
 	}
 
-	extracted.ClaimsSet = map[string]interface{}{
+	extracted := map[string]interface{}{
```
yogeshbdeshpande, line 60: as mentioned earlier, also, we can call this as `claimSet`

yogeshbdeshpande (approved):
Left some minor comments for you to have a look at! 

Otherwise LGTM!

thomas-fossati (approved):
awesome, thanks!

handler/istorehandler.go
```diff
@@ -14,11 +14,20 @@ import (
...
 
-	// GetTrustAnchorIDs returns an array of trust anchor identifiers used
+	// GetTrustAnchorIDs returns a slice of trust anchor identifiers used
 	// to retrieve the trust anchors associated with this token. The trust anchors may be necessary to validate the
 	// entire token and/or extract its claims (if it is encrypted).
 	GetTrustAnchorIDs(token *proto.AttestationToken) ([]string, error)
 
+	// GetRefValueIDs returns a slice of identifiers used to retrieve
+	// reference values for an attestation scheme, using the claims
+	// extracted from attestation token and the assoced trust anchors.
```
yogeshbdeshpande, line 24:
```suggestion
	// extracted from attestation token and the associated trust anchors.
```

Issue #207, opened 2023-12-13
Enhancement: Add a Store API to handle multiple IDs  in a single request, e.g. so that a Single Request example:  `o.EnStore.GetMultiple(appraisal.EvidenceContext.ReferenceIds)` can be more optimal


yogeshbdeshpande:
Note for the future (not necessary for this pull): Rather that looping over multiple requests to the store, we should add a Store API to handle multiple IDs  in a single request, e.g. so that his can just be `o.EnStore.GetMultiple(appraisal.EvidenceContext.ReferenceIds)`.

_Originally posted by @setrofim in https://github.com/veraison/services/pull/205#discussion_r1425121563_

setrofim:
No longer relevant post `corim-store` migration.

PR #220, opened 2024-04-30
kvstore: add GetMultiple

setrofim:
Add GetMultiple method to IKVStore that allows fetching values for multiple keys from a store with a single call. This simplifies the usage of the store inside trusted services and can potentially improve performance for remote store implementations where call overhead may be non-trivial.

This addresses https://github.com/veraison/services/issues/207

### DIFF
--- a/vts/trustedservices/trustedservices_grpc.go
+++ b/vts/trustedservices/trustedservices_grpc.go
@@ -393,19 +393,13 @@ func (o *GRPC) GetAttestation(
 		"software-id", appraisal.EvidenceContext.ReferenceIds,
 		"trust-anchor-id", appraisal.EvidenceContext.TrustAnchorIds)
 
-	var multEndorsements []string
-	for _, refvalID := range appraisal.EvidenceContext.ReferenceIds {
-
-		endorsements, err := o.EnStore.Get(refvalID)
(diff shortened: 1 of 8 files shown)


vts/trustedservices/trustedservices_grpc.go
```diff
@@ -393,35 +393,29 @@ func (o *GRPC) GetAttestation(
...
-	for _, refvalID := range appraisal.EvidenceContext.ReferenceIds {
-
-		endorsements, err := o.EnStore.Get(refvalID)
-		if err != nil && !errors.Is(err, kvstore.ErrKeyNotFound) {
-			return o.finalize(appraisal, err)
-		}
-
-		o.logger.Debugw("obtained endorsements", "endorsements", endorsements)
-		multEndorsements = append(multEndorsements, endorsements...)
+	endorsements, err := o.EnStore.GetMultiple(appraisal.EvidenceContext.ReferenceIds)
```
yogeshbdeshpande, line 396: @setrofim : Why the same API is not applicable to line 448-452 ?

yogeshbdeshpande:
Also, is the same not true for Add() Interface, used with one example on main branch:
`trustedservices_grpc.go`  line 318-322

thomas-fossati:
Just checking since I have lost the context: is this change still required?

PR #228, opened 2024-06-03
test(scheme): add missing GetRefValueIDs tests

setrofim:
When GetRefValueIDs was moved to StoreHandler, it was no longer tested as part of the existing ExtractClaims tests. Add StoreHandler tests to rectify this.


yogeshbdeshpande (approved):
Thank you for the change, it looks good to me!

Issue #191, opened 2023-09-05
TLS FTW!


thomas-fossati:
Authentication and integrity of the transport channels between clients and Veraison services, as well as between services, are prerequisite for the overall system security.  Encryption may also be a requisite in some (most) cases.

Currently:
* all Veraison REST APIs run over plain-text HTTP
* intra-service gRPC traffic is over plain TCP
* communication to extra services (e.g., KeyCloak) is over HTTP

We should ensure all traffic above is carried over secure (authenticated and encrypted) channels.

thomas-fossati:
> @thomas-fossati I would prefer acronym expansion in the title of the work

this is not an acronym, it's slang 😄

yogeshbdeshpande:
> > @thomas-fossati I would prefer acronym expansion in the title of the work
> 
> this is not an acronym, it's slang 😄

Then I would certainly change it!

setrofim:
It is both -- the acronym stands for "For The Win", which is a well-known net expression. And as a counter-point  I would leave it in :D -- it brings the point across nicely.

yogeshbdeshpande:
@setrofim It is not completely true. Google search reveals other terms also, not worth mentioning it here !

setrofim:
Sure, pretty much any TLA will have multiple decodings (a lot of which will not be worth mentioning). However, outside of specific contexts, FTW is broadly and unambiguously undestood to mean "For The Win".  (Just like "IIRC" can decode as "International Integrated Reporting Council" or "If I Really Cared", but is broadly understood to mean "If I Understand Correctly" in general).

thomas-fossati:
> @setrofim It is not completely true. Google search reveals other terms also, not worth mentioning it here !

LOL, the context is everything :-)

thomas-fossati:
The scope of this issue is wider than just the external API.  It includes:

* [x] all Veraison REST APIs run over plain-text HTTP
* [ ] intra-service gRPC traffic is over plain TCP
* [ ] communication to extra services (e.g., KeyCloak) is over HTTP

#229 addresses the first of the three points.

Alternatively, we could move the remaining two to new issues and mark this closed.

setrofim:
Resolved by https://github.com/veraison/services/pull/234

PR #229, opened 2024-06-06
feat: implement TLS support

setrofim:
Implement support for TLS for the API services. This introduces three new configurations to each service:

- protocol: this defaults to "https"; switching this to "http" diables TLS.
- cert: this is used to specify the path to the server certificate
- cert-key: this is use to specify the path to the key associated with the server certificate

Fixes https://github.com/veraison/services/issues/191


management/cmd/management-service/README.md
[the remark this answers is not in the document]
setrofim: > I think we want to make TLS the default. A user who wants to do plain-text HTTP should opt-out. I'd add a further configuration key insecure-api that such a user needs to set explicitly.

mmm that would be a breaking change for existing users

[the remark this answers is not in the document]
setrofim: i've implemented the change.

yogeshbdeshpande:
@setrofim Your commit comment needs to be corrected.

As of now it reads:

` - protocol: this defaults to "https"; switching this to "http" enables
  TLS.`

provisioning/cmd/provisioning-service/README.md
```diff
@@ -17,6 +17,9 @@ configuration:
 - `listen-addr` (optional): the address, in the form `<host>:<port>` the provisioning
   server will be listening on. If not specified, this defaults to
   `localhost:8888`.
+- `protocol` (optional): the protocol that will be used. Must be either "http" or "https". Defaults to "https" if not specified.
+- `cert` (optional): path to the x509 certificate to be used. Must be specified if protocol is "https"
```
yogeshbdeshpande, line 21: Should there be default certificates and path to default certificates, when the protocol is omitted and defaults to `https` ?

verification/cmd/verification-service/README.md
```diff
@@ -13,6 +13,9 @@ configuration:
 - `listen-addr` (optional): the address, in the form `<host>:<port>` the verification
   server will be listening on. If not specified, this defaults to
   `localhost:8080`.
+- `protocol` (optional): the protocol that will be used. Must be either "http" or "https". Defaults to "https" if not specified.
+- `cert` (optional): path to the x509 certificate to be used. Must be specified if protocol is "https"
```
yogeshbdeshpande, line 17: Similar question as `provisioning`

provisioning/cmd/provisioning-service/README.md
setrofim: No because there cannot be a sinsible default. Certs is something that _must_ be sourced by the user -- it's part of the deployment, not the code.

yogeshbdeshpande: Also it would be good to refer the new additions and `if required` any mandatory parameters to be supplied or defaults used in the README's associated with P, V, as below..
https://github.com/veraison/services/tree/main/provisioning

yogeshbdeshpande: > No because there cannot be a sensible default. Certs is something that _must_ be sourced by the user -- it's part of the deployment, not the code.

ok, understood, then I think, this does not become an `optional` parameter, as by default this is required, unless the user explicitly sets the protocol parameter to "http" to change the default behaviour. This aspect must be clarified in the README.

setrofim: Yeah good point. This was optional when the default was HTTP.

setrofim: actually the doc already clarifies that

verification/cmd/verification-service/README.md
```diff
@@ -13,6 +13,9 @@ configuration:
 - `listen-addr` (optional): the address, in the form `<host>:<port>` the verification
   server will be listening on. If not specified, this defaults to
   `localhost:8080`.
+- `protocol` (optional): the protocol that will be used. Must be either "http" or "https". Defaults to "https" if not specified.
```
yogeshbdeshpande, line 16:
```suggestion
- `protocol` (optional): the protocol that will be used. Defaults to "https" if not specified. Must be either "http" or "https".
```

yogeshbdeshpande (approved):
LGTM!

PR #231, opened 2024-06-12
fix: TLS for Keycloak and fix certs in integ tests

setrofim:
Enable TLS for Keycloak authentication (prior commit only enabled TLS for REST API).

This also fixes CA cert usage inside integration tests. Python requests ignores the system CA certs, using its own certificates that are bundled withe the package. It can be told to use the system CA cert bundle by setting REQUESTS_CA_BUNDLE environment variable.


thomas-fossati (approved):
LGTM, thanks!

yogeshbdeshpande (approved):
LGTM!

PR #232, opened 2024-06-14
fix: correctly report VTS down in API server logs

setrofim:
API frontends may be started before VTS. On start up, the frontend attempts to connect to the service and if the service is not ready, it is supposed to output a warning.

This was established by checking for error from
vtsClient.GetServiceState(). However, this will not always error. If a connection cannot be established this may return no error, but report the service as down.

In addition to checking for error, check that the service state is "READY", and warn if its not.


yogeshbdeshpande (approved):
LGTM! Thanks!

PR #233, opened 2024-06-17
Cca comb plugin 13 6

yogeshbdeshpande:


yogeshbdeshpande:
Issue #235 to address further work on Integration Test Cases...

scheme/psa-iot/psa_extractor.go
```diff
@@ -0,0 +1,158 @@
...
+)
+
+type PsaExtractor struct {
+	Profile string
+}
+
+func (o PsaExtractor) RefValExtractor(rv comid.ReferenceValue) ([]*handler.Endorsement, error) {
+	var classAttrs platform.ClassAttributes
+
+	if (o.Profile == "") || (o.Profile != "http://arm.com/psa/iot/1") {
```
setrofim: Can just be `if o.Profile != "http://arm.com/psa/iot/1"`

scheme/common/cca/realm/cca_realm_attributes.go
```diff
@@ -0,0 +1,111 @@
...
+	"errors"
+	"fmt"
+	"strings"
+
+	"github.com/veraison/corim/comid"
+	"github.com/veraison/services/log"
+)
+
+type RealmAttributes struct {
+	Rim       *[]byte
```
setrofim: Shouldn't the initialisms be capilatised, i.e. `RIM`, `REM`, `RPV` (to be consustent with things like `UUID`, `HashAlgID` etc..)?

scheme/common/cca/realm/cca_realm_extractor.go
```diff
@@ -0,0 +1,100 @@
...
+	"encoding/json"
+	"fmt"
+
+	"github.com/veraison/corim/comid"
+	"github.com/veraison/services/handler"
+)
+
+type CcaRealmExtractor struct {
+	Scheme    string
+	SubScheme string
```
setrofim: `SubScheme` not necessary?

scheme/common/cca/realm/realm_utils.go
```diff
@@ -0,0 +1,96 @@
...
+	"errors"
+	"fmt"
+	"net/url"
+	"strings"
+
+	"github.com/veraison/ccatoken"
+	"github.com/veraison/services/log"
+)
+
+func GetRim(subscheme string, attr json.RawMessage) (string, error) {
```
setrofim: As with field names, these should probably be `GetRIM`, `GetRPV`, etc...

scheme/common/arm/handlers.go
```diff
@@ -10,47 +10,68 @@ import (
...
-	Label  string `cca:"CCA_SSD_PLATFORM.platform-config-label" parcca:"PARSEC_CCA.platform-config-label"`
-	Value  []byte `cca:"CCA_SSD_PLATFORM.platform-config-id" parcca:"PARSEC_CCA.platform-config-id"`
+	ImplID []byte `json:"impl-id"`
+	Model  string `json:"hw-model"`
+	Vendor string `json:"hw-vendor"`
+	Label  string `json:"platform-config-label"`
+	Value  []byte `json:"platform-config-id"`
 }
 
 func SynthKeysFromRefValue(scheme string, tenantID string,
```
setrofim: There is no point in having this method in common code, as it is just  a switch on scheme, and so is not actually common code.

Instead  implement a `SynthKeysFromRefValue` for each of the shcems bleow that justs calls the relevant  function.

scheme/common/arm/handlers.go
```diff
@@ -60,7 +81,58 @@ func SynthKeysFromRefValue(scheme string, tenantID string,
 	log.Debugf("Scheme %s Plugin Reference Value Look Up Key= %s\n", scheme, lookupKey)
 
 	return []string{lookupKey}, nil
+}
+
+func GetReferenceIDs(
```
setrofim: Ditto.

scheme/arm-cca/evidence_handler.go
```diff
@@ -10,8 +10,7 @@ import (
 	"fmt"
 
 	"github.com/veraison/ccatoken"
-	"github.com/veraison/ear"
-	"github.com/veraison/psatoken"
+	ar "github.com/veraison/ear"
```
setrofim: Do not alias packge unless absolutely necessary. This makes it harder to read code. When I'm reading a function, I know what an `ear.AttestationResult` is, I won't know whan an `ar.AttestationResult`, and will have to just to the imports to resolve the alias.

scheme/arm-cca/platform.go
```diff
@@ -0,0 +1,79 @@
...
+
+	"github.com/veraison/ear"
+	"github.com/veraison/psatoken"
+	"github.com/veraison/services/handler"
+	"github.com/veraison/services/log"
+	"github.com/veraison/services/scheme/common"
+	"github.com/veraison/services/scheme/common/arm"
+)
+
+type Cca_platform_attester struct {
```
setrofim, line 17: Should be `CCAPlatformAttester`

scheme/arm-cca/realm.go
```diff
@@ -0,0 +1,119 @@
...
+
+	"github.com/veraison/ccatoken"
+	"github.com/veraison/ear"
+	"github.com/veraison/services/handler"
+	"github.com/veraison/services/log"
+	"github.com/veraison/services/scheme/common/arm"
+	realm "github.com/veraison/services/scheme/common/cca/realm"
+)
+
+type Cca_realm_attester struct {
```
setrofim, line 19: Should be `CCARealmAttester`

setrofim (changes requested):

scheme/common/arm/handlers.go
yogeshbdeshpande: @setrofim : May be you are seeing the old code ? Can you please check, again, this method no longer exists! It is just in the squashing sequence, you may be seeing this ?
Only method exposed is: `SynthKeysForPlatform` & `GetPlatformReferenceIDs` which does not have any scheme specific code changes!

setrofim: I was reviewing one commit at a time, it's possilbe this got replaced by a later commit..

yogeshbdeshpande: Sure, this has been addressed and I have rechecked it again today!

scheme/common/arm/handlers.go
yogeshbdeshpande: Similarly, there is no `GetReferenceIDs()` now only `GetPlatformReferenceIDs()` shared across all the Arm platforms

scheme/arm-cca/evidence_handler.go
yogeshbdeshpande: Thank you for your comment, This specific one on `ear` is addressed once I noticed a similar comment on the other PR. It is now used only when it is `absolutely` necessary! Resolving this comment, as it is addressed!

scheme/common/cca/realm/cca_realm_extractor.go
yogeshbdeshpande: Has been addressed in the later commits, the `SubScheme` was removed subsequently!

scheme/psa-iot/psa_extractor.go
yogeshbdeshpande: Make sense, will do the correction!

scheme/common/cca/realm/cca_realm_attributes.go
yogeshbdeshpande: Will do it now!

scheme/common/cca/realm/realm_utils.go
yogeshbdeshpande: Now have capitalized RIM, REM and RPV in the code base.

scheme/psa-iot/psa_extractor.go
yogeshbdeshpande: Done!

setrofim (approved):
LGTM!

scheme/arm-cca/corim_extractor.go
```diff
@@ -0,0 +1,46 @@
...
+	}
+}
+
+func (o CorimExtractor) TaExtractor(avk comid.AttestVerifKey) (*handler.Endorsement, error) {
+	switch o.Profile {
+	case "http://arm.com/cca/ssd/1":
+		subScheme := &platform.CcaSsdExtractor{Scheme: SchemeName}
+		return subScheme.TaExtractor(avk)
+	case "http://arm.com/cca/realm/1":
+		return nil, fmt.Errorf("incorrect Trust Anchor Extractor invoked for Profile: %s", o.Profile)
```
thomas-fossati, line 38: This can be removed: there are no trust anchors in realms, and we already have the catch-all below.

scheme/arm-cca/corim_extractor.go
```diff
@@ -0,0 +1,46 @@
...
+	default:
+		return nil, fmt.Errorf("invalid profile: %s for Scheme ARM_CCA ", o.Profile)
```
thomas-fossati, line 40:
```suggestion
		return nil, fmt.Errorf("invalid profile %s for scheme %s", o.Profile, SchemeName)
```

scheme/arm-cca/evidence_handler.go
```diff
@@ -117,84 +115,48 @@ func (s EvidenceHandler) AppraiseEvidence(
...
-
-func populateAttestationResult(
-	result *ear.AttestationResult,
-	evidence map[string]interface{},
-	endorsements []handler.Endorsement,
-) error {
-	claims, err := common.MapToClaims(evidence["platform"].(map[string]interface{}))
+	var err error
+	var status ear.TrustTier
+	endorsements, err = getEndorsementsFromString(endorsementsStrings)
```
thomas-fossati, line 120: Are the variable declarations for `endorsements` and `err` really needed?

Why not instead:

```suggestion
	endorsements, err := getEndorsementsFromString(endorsementsStrings)
```

scheme/arm-cca/evidence_handler.go
```diff
@@ -117,84 +115,48 @@ func (s EvidenceHandler) AppraiseEvidence(
...
-		appraisal.TrustVector.StorageOpaque = ear.HwKeysEncryptedSecretsClaim
-	} else {
-		appraisal.TrustVector.InstanceIdentity = ear.UntrustworthyInstanceClaim
-		appraisal.TrustVector.RuntimeOpaque = ear.VisibleMemoryRuntimeClaim
-		appraisal.TrustVector.StorageOpaque = ear.UnencryptedSecretsClaim
+	claims = evidence["realm"].(map[string]interface{})
+	realmappraisal := &ear.Appraisal{
+		TrustVector: &ear.TrustVector{},
+		Status:      &status,
 	}
```
thomas-fossati, line 142: Not sure why you need to do this, wouldn't it be sufficient to create a default object?

```suggestion
	realmappraisal := &ear.Appraisal{}
```

(While you are here, maybe use camel-case for the variable name: `realmAppraisal`)

scheme/arm-cca/evidence_handler_test.go
```diff
@@ -93,11 +92,75 @@ func Test_AppraiseEvidence_mismatch_refval_cfg(t *testing.T) { // nolint: dupl
...
 
+func Test_AppraiseEvidence_Realm(t *testing.T) { // nolint: dupl
+
+	tvs := []struct {
+		desc           string
+		input          string
+		expectedStatus ear.TrustTier
+		expectedExec   ear.TrustClaim
+	}{
+
```
thomas-fossati, line 103: spurious NL

```suggestion
```

scheme/arm-cca/evidence_handler_test.go
```diff
@@ -93,11 +92,75 @@ func Test_AppraiseEvidence_mismatch_refval_cfg(t *testing.T) { // nolint: dupl
 	assert.Equal(t, attestation.TrustVector.Configuration, ear.UnsafeConfigClaim)
 }
 
+func Test_AppraiseEvidence_Realm(t *testing.T) { // nolint: dupl
+
```
thomas-fossati, line 96: spurious NL

```suggestion
```

scheme/arm-cca/evidence_handler_test.go
```diff
@@ -93,11 +92,75 @@ func Test_AppraiseEvidence_mismatch_refval_cfg(t *testing.T) { // nolint: dupl
...
+		expectedExec   ear.TrustClaim
+	}{
+
+		{
+			desc:           "No realm endorsements",
+			input:          "test/realm/no-realm-endorsements.json",
+			expectedStatus: ear.TrustTierWarning,
+			expectedExec:   ear.UnrecognizedRuntimeClaim,
+		},
+
```
thomas-fossati, line 110: bunch of spurious NLs her and below
```suggestion
```

thomas-fossati (commented):
some half-way through comments

yogeshbdeshpande:
Thank You @thomas-fossati for your comments, look forward to see the remaining before making the changes!

scheme/arm-cca/evidence_handler.go
yogeshbdeshpande: > Not sure why you need to do this, wouldn't it be sufficient to create a default object?
> 
> (While you are here, maybe use camel-case for the variable name: `realmAppraisal`)

No this is required otherwise just creating the way you suggested would trap reason `nil` TrustVector

yogeshbdeshpande: I will change to camel-case instead ..

thomas-fossati: > > Not sure why you need to do this, wouldn't it be sufficient to create a default object?
> > (While you are here, maybe use camel-case for the variable name: `realmAppraisal`)
> 
> No this is required otherwise just creating the way you suggested would trap reason `nil` TrustVector

Sorry, I don't get the logic.  Is `PerformAppraisal` using `realmAppraisal` as an in-out parameter?  If not - as I'd guess - just create the object in it and return it, no need to pass it as a parameter, no?

integration-tests/data/results/cca.freshness-fail.json
```diff
@@ -1,18 +1,20 @@
...
+                "runtime-opaque": 99,
+                "sourced-data": 99,
+                "storage-opaque": 99
+            },
+            "ear.veraison.policy-claims": {
+                "problem": "integrity validation failed"
+            }
+    }
 
+}
```
thomas-fossati, line 20: Indentation has gone a bit wild :-)

Can you please do a `jq` sweep over the JSON files you have touched?

Something like:
```sh
git diff --name-only main \
  | grep "\.json$" \
  | while read f ; do [ -f $f ] && ( jq . $f | sponge $f ) ; done
```
should do the trick.

(Note that to run the above you may need to install `sponge(1)`.  On Debian/Ubuntu, do `sudo apt-get install moreutils`.)

thomas-fossati: PS: also worth going through the READMEs that you have touched to make sure that any JSON example is properly formatted.

scheme/arm-cca/platform.go
```diff
@@ -0,0 +1,79 @@
...
+		appraisal.TrustVector.Configuration = ear.UnsafeConfigClaim
+		log.Debug("matchPlatformConfig Failed")
+	}
+
+	appraisal.UpdateStatusFromTrustVector()
+
+	appraisal.VeraisonAnnotatedEvidence = &claimsMap
+
+	return nil
+}
```
thomas-fossati, line 79: Related to https://github.com/veraison/services/pull/233/files#r1644911152 I suggest we make `PerformAppraisal` return the appraisal object:

```suggestion
func (CCAPlatformAttester) PerformAppraisal(
	claimsMap map[string]interface{},
	endorsements []handler.Endorsement,
) (*ear.Appraisal, error) {
	claims, err := common.MapToClaims(claimsMap)
	if err != nil {
		return nil, fmt.Errorf("unable to get platform claims from platform claims map: %w", err)
	}

    trustVector := ear.TrustVector{}
	// once the signature on the token is verified, we can claim the HW is
	// authentic
	trustVector.Hardware = ear.GenuineHardwareClaim

	rawLifeCycle, err := claims.GetSecurityLifeCycle()
	if err != nil {
		return nil, handler.BadEvidence(err)
	}

	lifeCycle := psatoken.CcaLifeCycleToState(rawLifeCycle)
	if lifeCycle == psatoken.CcaStateSecured ||
		lifeCycle == psatoken.CcaStateNonCcaPlatformDebug {
		trustVector.InstanceIdentity = ear.TrustworthyInstanceClaim
		trustVector.RuntimeOpaque = ear.ApprovedRuntimeClaim
		trustVector.StorageOpaque = ear.HwKeysEncryptedSecretsClaim
	} else {
		trustVector.InstanceIdentity = ear.UntrustworthyInstanceClaim
		trustVector.RuntimeOpaque = ear.VisibleMemoryRuntimeClaim
		trustVector.StorageOpaque = ear.UnencryptedSecretsClaim
	}

	swComps := arm.FilterRefVal(endorsements, "sw-component")
	match := arm.MatchSoftware(SchemeName, claims, swComps)
	if match {
		trustVector.Executables = ear.ApprovedRuntimeClaim
	} else {
		appraisal.TrustVector.Executables = ear.UnrecognizedRuntimeClaim
	}

	platformConfig := arm.FilterRefVal(endorsements, "platform-config")
	match = arm.MatchPlatformConfig(SchemeName, claims, platformConfig)

	if match {
		trustVector.Configuration = ear.ApprovedConfigClaim
	} else {
		appraisal.TrustVector.Configuration = ear.UnsafeConfigClaim
	}

    appraisal := ear.Appraisal{
        TrustVector: trustVector,
    }

	appraisal.UpdateStatusFromTrustVector()
	appraisal.VeraisonAnnotatedEvidence = &claimsMap

	return &appraisal, nil
}
```

scheme/arm-cca/plugin/combined/Makefile
```diff
@@ -1,8 +1,8 @@
 # Copyright 2024 Contributors to the Veraison project.
 # SPDX-License-Identifier: Apache-2.0
 
-PLUGIN := ../../../bin/cca-store-handler.plugin
-GOPKG := github.com/veraison/services/scheme/cca-ssd-platform
+PLUGIN := ../../../bin/arm-cca.plugin
+GOPKG := GOPKG := github.com/veraison/services/scheme/arm-cca
```
thomas-fossati, line 5: Not sure how that managed to work!  I bow to the superior self-healing properties of GNU make :)

```suggestion
GOPKG := github.com/veraison/services/scheme/arm-cca
```

scheme/arm-cca/plugin/endorsement-handler/Makefile
```diff
@@ -1,8 +1,8 @@
-# Copyright 2021 Contributors to the Veraison project.
+# Copyright 2024 Contributors to the Veraison project.
```
thomas-fossati, line 1: just a nit, shouldn't this be 2021-2024?

scheme/common/cca/platform/cca_ssd_extractor.go
```diff
@@ -1,6 +1,6 @@
-// Copyright 2022-2024 Contributors to the Veraison project.
+// Copyright 2024 Contributors to the Veraison project.
 // SPDX-License-Identifier: Apache-2.0
-package arm
+package cca
```
thomas-fossati, line 3: The convention for naming packages (see https://go.dev/doc/effective_go#package-names) is that _"[the package] name is the base name of its source directory"_.  So this should be `platform` rather than `cca`.

scheme/arm-cca/realm.go
```diff
@@ -0,0 +1,119 @@
...
+	"bytes"
+	"encoding/base64"
+	"fmt"
+
+	"github.com/veraison/ccatoken"
+	"github.com/veraison/ear"
+	"github.com/veraison/services/handler"
+	"github.com/veraison/services/log"
+	"github.com/veraison/services/scheme/common/arm"
+	realm "github.com/veraison/services/scheme/common/cca/realm"
```
thomas-fossati, line 16: is the alias needed?

scheme/arm-cca/corim_extractor.go
```diff
@@ -0,0 +1,46 @@
...
+// SPDX-License-Identifier: Apache-2.0
+package arm_cca
+
+import (
+	"fmt"
+
+	"github.com/veraison/corim/comid"
+	"github.com/veraison/services/handler"
+	platform "github.com/veraison/services/scheme/common/cca/platform"
+	realm "github.com/veraison/services/scheme/common/cca/realm"
```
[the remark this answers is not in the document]
thomas-fossati: See https://github.com/veraison/services/pull/233/files#r1645622827

scheme/arm-cca/realm.go
thomas-fossati: see https://github.com/veraison/services/pull/233/files#r1645622827

scheme/arm-cca/evidence_handler.go
```diff
@@ -117,84 +115,48 @@ func (s EvidenceHandler) AppraiseEvidence(
...
+	var status ear.TrustTier
+	endorsements, err = getEndorsementsFromString(endorsementsStrings)
 	if err != nil {
-		return err
+		return nil, err
 	}
+	var claims map[string]interface{}
+	CcaPlatform := &CCAPlatformAttester{}
+	CcaRealm := &CCARealmAttester{}
+	evidence := ec.Evidence.AsMap()
```
thomas-fossati, line 127: Move these near their usage site.

scheme/arm-cca/evidence_handler.go
```diff
@@ -117,84 +115,48 @@ func (s EvidenceHandler) AppraiseEvidence(
...
-) error {
-	claims, err := common.MapToClaims(evidence["platform"].(map[string]interface{}))
+	var err error
+	var status ear.TrustTier
+	endorsements, err = getEndorsementsFromString(endorsementsStrings)
 	if err != nil {
-		return err
+		return nil, err
 	}
+	var claims map[string]interface{}
```
thomas-fossati, line 124: This can be defined directly at L132 (`:=`)

scheme/arm-cca/realm.go
```diff
@@ -0,0 +1,119 @@
...
+)
+
+type CCARealmAttester struct {
+}
+
+func (CCARealmAttester) PerformAppraisal(
```
thomas-fossati, line 22: ditto here: https://github.com/veraison/services/pull/233/files#r1644911152

scheme/arm-cca/realm.go
```diff
@@ -0,0 +1,119 @@
...
+	appraisal.VeraisonAnnotatedEvidence = &claimsMap
+
+	return nil
+}
+
+func matchRimRpv(claims ccatoken.IClaims, endorsement *handler.Endorsement) bool {
+	// Compare RIM and PV (if provided)
+	rimClaim, err := claims.GetInitialMeasurement()
+	if err != nil {
+		log.Errorf("matchRimRpv failed: %w", handler.BadEvidence(err))
```
thomas-fossati, line 63: There is no reason to wrap the error in `BadEvidence`

scheme/arm-cca/realm.go
```diff
@@ -0,0 +1,119 @@
...
+		return false
+	}
+	pvClaim, err := claims.GetPersonalizationValue()
+	if err != nil {
+		log.Errorf("matchRimRpv failed: %w", handler.BadEvidence(err))
```
thomas-fossati, line 68: ditto

scheme/arm-cca/realm.go
```diff
@@ -0,0 +1,119 @@
...
+			appraisal.TrustVector.Executables = ear.ContraindicatedRuntimeClaim
+		}
+	}
+	appraisal.UpdateStatusFromTrustVector()
+	appraisal.VeraisonAnnotatedEvidence = &claimsMap
+
+	return nil
+}
+
+func matchRimRpv(claims ccatoken.IClaims, endorsement *handler.Endorsement) bool {
```
thomas-fossati, line 59: Any specific reason to merge RIM and PV matching in the same function?

scheme/arm-cca/realm.go
thomas-fossati, line 63: These log messages do not seem to provide the information needed to correlate the error with the current appraisal.  Which makes them not very useful, apart from when debugging an isolated transaction.

scheme/arm-cca/realm.go
yogeshbdeshpande: Yes, if the Rim matches but RPV (if provided in Reference VALUE) does not match, then we could not identify the correct Realm instance correctly, hence they are matched together!

scheme/arm-cca/realm.go
```diff
@@ -0,0 +1,119 @@
...
+		}
+	}
+	return false
+}
+
+func matchREMs(claims ccatoken.IClaims, endorsement *handler.Endorsement) bool {
+	remMatch := false
+	remsClaim, err := claims.GetExtensibleMeasurements()
+	if err != nil {
+		log.Errorf("unable to get realm extensible measurements from claims: %w", handler.BadEvidence(err))
```
thomas-fossati, line 100: no need to wrap `err`

integration-tests/data/results/cca.freshness-fail.json
yogeshbdeshpande: Thanks, Totally agree with you, I will run `jq` with all impacted `json` files and update the same!

scheme/arm-cca/realm.go
thomas-fossati: Hmm, but that's not a necessity.  You could do:

```golang
func matchRIM(actual []byte, endorsement *handler.Endorsement) error {
    expected, err := realm.GetRim(endorsement.Attributes)
    if err != nil {
        return err
    }

    return bytes.Equal(actual, expected)
}

func matchRPV(actual []byte, endorsement *handler.Endorsement) error {
    expected, err := realm.GetRpv(endorsement.Attributes)
    if err != nil {
        if errors.Is(err, ErrNotFound) {
            return nil
        }
        return err
    }

    return bytes.Equal(actual, expected)
}
```
and then in the calling site:
```golang
    if matchRPV(rpv, endorsement) != nil && matchREM(rem, endorsement) != nil {
        match = true
    }
```

yogeshbdeshpande: True that, but at the end of the day they both are `intended` to identify `uniqueness of a Realm` , so logically compared together!

scheme/arm-cca/store_handler.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+func (s StoreHandler) SynthKeysFromRefValue(
+	tenantID string,
+	refVal *handler.Endorsement,
+) ([]string, error) {
+	switch refVal.SubType {
+	case "sw-component", "platform-config":
+		return arm.SynthKeysForPlatform(SchemeName, tenantID, refVal)
+	case "realm.reference-value":
```
thomas-fossati, line 37: naming is a bit wild here:

* non-namespaced: `sw-component`
* namespaced with `-` separator: `platform-config`
* namespaced with `.` separator: `realm.reference-value`

:-)

probably worth fixing?

scheme/arm-cca/realm.go
yogeshbdeshpande: Also RPV is Optional so in absence of RPV  `matchRPV` may fail, but that does not mean anything, so the comparison has to be in correct context, hence best to be done together!

scheme/arm-cca/evidence_handler.go
```diff
@@ -117,84 +115,48 @@ func (s EvidenceHandler) AppraiseEvidence(
...
-	if lifeCycle == psatoken.CcaStateSecured ||
-		lifeCycle == psatoken.CcaStateNonCcaPlatformDebug {
-		appraisal.TrustVector.InstanceIdentity = ear.TrustworthyInstanceClaim
-		appraisal.TrustVector.RuntimeOpaque = ear.ApprovedRuntimeClaim
-		appraisal.TrustVector.StorageOpaque = ear.HwKeysEncryptedSecretsClaim
-	} else {
-		appraisal.TrustVector.InstanceIdentity = ear.UntrustworthyInstanceClaim
-		appraisal.TrustVector.RuntimeOpaque = ear.VisibleMemoryRuntimeClaim
-		appraisal.TrustVector.StorageOpaque = ear.UnencryptedSecretsClaim
+	claims = evidence["realm"].(map[string]interface{})
```
thomas-fossati, line 138: this type assertion may fail

scheme/arm-cca/store_handler.go
```diff
@@ -0,0 +1,85 @@
...
+) ([]string, error) {
+	platformClaimsMap, ok := claims["platform"].(map[string]interface{})
+	if !ok {
+		return nil, fmt.Errorf("claims do not contain platform map: %v", claims)
+	}
+	pids, err := arm.GetPlatformReferenceIDs(SchemeName, tenantID, platformClaimsMap)
+	if err != nil {
+		return nil, fmt.Errorf("unable to get cca platform reference IDs: %w", err)
+	}
+	rids, err := realm.GetRealmReferenceIDs(SchemeName, tenantID, claims)
```
thomas-fossati, line 80: Why is the realm looking at the whole claims-set rather than just the realm part?

scheme/arm-cca/store_handler.go
```diff
@@ -0,0 +1,85 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("unable to get cca realm reference IDs: %w", err)
+	}
+	return append(pids, rids...), nil
```
thomas-fossati, line 84: Question: how can the caller tell a platform ID from a realm ID?

scheme/arm-cca/store_handler.go
yogeshbdeshpande: Yes, this should mirror how platform part is implemented, I shall change this, to be more clear!

scheme/common/arm/handlers.go
```diff
@@ -60,7 +56,24 @@ func SynthKeysFromRefValue(scheme string, tenantID string,
...
 
 	return []string{lookupKey}, nil
+}
+
+func GetPlatformReferenceIDs(
+	scheme string,
+	tenantID string,
+	claims map[string]interface{},
+) ([]string, error) {
+
```
thomas-fossati, line 66: spurious NL

scheme/arm-cca/store_handler.go
yogeshbdeshpande: Sorry, I did not get the question ? Perhaps clarify it better :)

scheme/common/cca/platform/cca_ssd_extractor.go
```diff
@@ -81,44 +70,41 @@ func (o Extractor) RefValExtractor(rv comid.ReferenceValue) ([]*handler.Endorsem
 	return refVals, nil
 }
 
-func extractMeasurement(
-	obj MeasurementExtractor,
+func (o CcaSsdExtractor) extractMeas(
```
thomas-fossati, line 73: no reason to shorten the function name.  Please keep it as it was before.

scheme/arm-cca/plugin/combined/Makefile
yogeshbdeshpande: Ah, good catch: I shall correct this!

scheme/common/cca/realm/realm_attributes.go
```diff
@@ -0,0 +1,115 @@
...
+		log.Debug("realm personalization value not present")
+		return nil
+	}
+	rpv, err := r.GetBytes()
+	if err != nil {
+		return err
+	} else if len(rpv) != 64 {
+		{
+			return fmt.Errorf("invalid length %d, for realm personalization value", len(rpv))
+		}
```
thomas-fossati, line 100: extra parenthesis:

```suggestion
		return fmt.Errorf("invalid length %d, for realm personalization value", len(rpv))
```

scheme/common/cca/realm/realm_attributes.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+	o.RPV = &rpv
+	return nil
+}
+
+func (o *RealmAttributes) Valid() error {
+	if o == nil {
+		return errors.New("no realm attributes object")
+	}
```
thomas-fossati, line 109: If this happen, the caller deserves to `panic` :-)

scheme/common/cca/realm/realm_classattributes.go
```diff
@@ -0,0 +1,51 @@
...
+	if classID == nil {
+		log.Debug("no classID in the environment")
+	} else {
+		UUID, err := classID.GetUUID()
+		if err != nil {
+			return fmt.Errorf("could not extract uu-id from class-id: %w", err)
+		}
+
+		if err := UUID.Valid(); err != nil {
+			return fmt.Errorf("no valid uu-id: %w", err)
```
thomas-fossati, line 38:
```suggestion
			return fmt.Errorf("no valid uuid: %w", err)
```

scheme/common/cca/realm/realm_classattributes.go
```diff
@@ -0,0 +1,51 @@
...
+	}
+
+	classID := class.ClassID
+
+	if classID == nil {
+		log.Debug("no classID in the environment")
+	} else {
+		UUID, err := classID.GetUUID()
+		if err != nil {
+			return fmt.Errorf("could not extract uu-id from class-id: %w", err)
```
thomas-fossati, line 34:
```suggestion
			return fmt.Errorf("could not extract uuid from class-id: %w", err)
```

scheme/common/cca/realm/realm_classattributes.go
```diff
@@ -0,0 +1,51 @@
...
+	UUID   *string
+	Vendor *string
+}
+
+// extract class variables from environment
+func (o *RealmClassAttributes) FromEnvironment(e comid.Environment) error {
+	class := e.Class
+
+	if class == nil {
+		log.Debug("no class in the environment")
```
thomas-fossati, line 23: these log messages look like debug printf's to me.  I suggest to remove them.

scheme/common/cca/realm/realm_extractor.go
```diff
@@ -0,0 +1,99 @@
...
+
+	if err := instAttrs.FromEnvironment(rv.Environment); err != nil {
+		return nil, fmt.Errorf("could not extract Realm instance attributes: %w", err)
+	}
+
+	// Measurements are encoded in a measurement-map of a CoMID
+	// reference-triple-record.  Since a measurement-map can encode one or more
+	// measurements, a single reference-triple-record can carry as many
+	// measurements as needed. Hence for a Realm Instance, all the measurements
+	// which comprise of both the "RIM" & "REM" measurements are carried in an
```
thomas-fossati, line 33:
```suggestion
	// which comprise both the "RIM" & "REM" measurements are carried in an
```

thomas-fossati, line 33: It's difficult to grasp the point of this comment. Please condense it to convey the essential information needed to understand the code.

scheme/common/cca/realm/realm_handlers.go
```diff
@@ -0,0 +1,45 @@
...
+
+func SynthKeysForCcaRealm(scheme string, tenantID string,
+	refVal *handler.Endorsement,
+) ([]string, error) {
+	rim, err := GetRIM(refVal.Attributes)
+	if err != nil {
+		return nil, fmt.Errorf("unable to get rim %w", err)
+	}
+	lookupKey := RealmRefValLookupKey(scheme, tenantID, rim)
+	log.Debugf("Scheme %s Plugin Reference Value Look Up Key= %s\n", scheme, lookupKey)
```
thomas-fossati, line 21: consider dropping this line

scheme/common/cca/realm/realm_instanceattributes.go
```diff
@@ -0,0 +1,29 @@
...
+import (
+	"errors"
+
+	"github.com/veraison/corim/comid"
+)
+
+type RealmInstanceAttributes struct {
+	InstID string
+}
+
```
thomas-fossati, line 14: Question: what is the semantics of this?

scheme/common/cca/realm/realm_utils.go
```diff
@@ -0,0 +1,96 @@
...
+
+	"github.com/veraison/ccatoken"
+	"github.com/veraison/services/log"
+)
+
+func GetRIM(attr json.RawMessage) (string, error) {
+	var at map[string]interface{}
+	err := json.Unmarshal(attr, &at)
+	if err != nil {
+		return "", fmt.Errorf("unable to get Instance ID: %w", err)
```
thomas-fossati, line 21:
```suggestion
		return "", fmt.Errorf("unable to get RIM: %w", err)
```

scheme/common/cca/realm/realm_handlers.go
```diff
@@ -0,0 +1,45 @@
...
+
+func GetRealmReferenceIDs(
+	scheme string,
+	tenantID string,
+	claims map[string]interface{},
+) ([]string, error) {
+	realmClaimsMap, ok := claims["realm"].(map[string]interface{})
+	if !ok {
+		return nil, fmt.Errorf("claims do not contain realm map: %v", claims)
+	}
```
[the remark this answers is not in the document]
yogeshbdeshpande: `Yes` we agreed to correct this, when you reviewed the store part of the same!

scheme/common/cca/realm/realm_utils.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+	key := "realm-initial-measurement"
+	rim, ok := at[key].(string)
```
thomas-fossati, line 24: just do:
```suggestion
	rim, ok := at["realm-initial-measurement"].(string)
```

scheme/common/cca/realm/realm_utils.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+func GetRPV(attr json.RawMessage) ([]byte, error) {
+	var at map[string]interface{}
+	err := json.Unmarshal(attr, &at)
+	if err != nil {
+		return nil, err
+	}
+	key := "realm-personalization-value"
+	r, ok := at[key].(string)
```
thomas-fossati, line 38: ditto

scheme/common/cca/realm/realm_utils.go
```diff
@@ -0,0 +1,96 @@
...
+	if !ok {
+		return nil, nil
+	}
```
thomas-fossati, line 41: I don't think this is correct.  This test would fail also incase PV is supplied but incorrectly encoded.  And you want to signal this as an error, I think.

There are two conditions to entangled here:

```golang
iv := at["realm-personalization-value]
// key missing => OK
if iv == nil {
    return nil, nil
}
// key not correctly encoded => !OK
v, ok := iv.(string)
if !ok {
    return fmt.Errorf(...)
}
```

scheme/common/cca/realm/realm_utils.go
```diff
@@ -0,0 +1,96 @@
...
+func GetREMs(attr json.RawMessage) ([][]byte, error) {
+	var at map[string]interface{}
+	var rems [][]byte
+	keys := []string{"rem0", "rem1", "rem2", "rem3"}
+	err := json.Unmarshal(attr, &at)
+	if err != nil {
+		return nil, err
+	}
+	for _, key := range keys {
+		log.Debugf("Rem key = %s", key)
```
thomas-fossati, line 58: please remove this

scheme/common/cca/realm/realm_utils.go
```diff
@@ -0,0 +1,96 @@
...
+		rem, ok := at[key].(string)
+		if ok {
+			brem, err := base64.StdEncoding.DecodeString(rem)
+			if err != nil {
+				return nil, err
+			}
+			rems = append(rems, brem)
+		} else {
+			log.Debugf("No Rem with key = %s", key)
+			break
```
thomas-fossati, line 68: Shouldn't this be an error?

scheme/common/unsignedcorim_decoder.go
```diff
@@ -41,10 +42,12 @@ func UnsignedCorimDecoder(
 			return nil, fmt.Errorf("found multiple profiles (expected exactly one): %s", strings.Join(profiles, ", "))
 		}
 		p := (*uc.Profiles)[0]
-		_, err := p.Get()
+		profile, err := p.Get()
+		log.Debugf("profile: %s", profile)
```
thomas-fossati, line 46: drop this line

scheme/common/utils.go
```diff
@@ -40,25 +41,27 @@ func MapToClaims(in map[string]interface{}) (psatoken.IClaims, error) {
 
 func GetImplID(scheme string, attr json.RawMessage) (string, error) {
 	var at map[string]interface{}
+
 	err := json.Unmarshal(attr, &at)
 	if err != nil {
-		return "", fmt.Errorf("unable to get Implementation ID: %w", err)
+		return "", fmt.Errorf("for scheme %s, unable to get Implementation ID for scheme: %w", scheme, err)
```
thomas-fossati, line 47:
```suggestion
		return "", fmt.Errorf("unable to get Implementation ID for scheme %s: %w", scheme, err)
```

scheme/common/utils.go
```diff
@@ -40,25 +41,27 @@ func MapToClaims(in map[string]interface{}) (psatoken.IClaims, error) {
...
 	}
-	key := scheme + ".impl-id"
+	key := "impl-id"
 	implID, ok := at[key].(string)
 	if !ok {
-		return "", errors.New("unable to get Implementation ID")
+		return "", fmt.Errorf("for scheme %s, unable to get Implementation ID", scheme)
```
thomas-fossati, line 52: ditto

scheme/common/utils.go
```diff
@@ -40,25 +41,27 @@ func MapToClaims(in map[string]interface{}) (psatoken.IClaims, error) {
...
 	}
 	return implID, nil
 }
 
 func GetInstID(scheme string, attr json.RawMessage) (string, error) {
 	var at map[string]interface{}
+	log.Debugf("get instance id for scheme: %s", scheme)
```
thomas-fossati, line 59: drop this

scheme/parsec-cca/parsec_cca_extractor.go
```diff
@@ -0,0 +1,36 @@
...
+	"github.com/veraison/services/handler"
+	ccaplatform "github.com/veraison/services/scheme/common/cca/platform"
+)
+
+type ParsecCcaExtractor struct {
+	Profile string
+}
+
+func (o ParsecCcaExtractor) RefValExtractor(rv comid.ReferenceValue) ([]*handler.Endorsement, error) {
+
```
thomas-fossati, line 18: spurious NL
```suggestion
```

scheme/parsec-cca/parsec_cca_extractor.go
```diff
@@ -0,0 +1,36 @@
...
+	if o.Profile != "tag:github.com/parallaxsecond,2023-03-03:cca" {
+		return nil, fmt.Errorf("invalid profile: %s, for Scheme: PARSEC_CCA", o.Profile)
```
thomas-fossati, line 20: extra comma
```suggestion
		return nil, fmt.Errorf("invalid profile: %s for scheme PARSEC_CCA", o.Profile)
```

scheme/parsec-cca/parsec_cca_extractor.go
```diff
@@ -0,0 +1,36 @@
...
+	}
+	subScheme := &ccaplatform.CcaSsdExtractor{}
+	return subScheme.RefValExtractor(rv)
+}
+
+func (o ParsecCcaExtractor) TaExtractor(avk comid.AttestVerifKey) (*handler.Endorsement, error) {
+	if o.Profile != "tag:github.com/parallaxsecond,2023-03-03:cca" {
+		return nil, fmt.Errorf("invalid profile: %s, for Scheme: PARSEC_CCA", o.Profile)
```
thomas-fossati, line 28: ditto

scheme/arm-cca/realm.go
thomas-fossati: > True that, but at the end of the day they both are intended to identify uniqueness of a Realm , so logically compared together!

I don't see it as a logical necessity.  Also, my point is it makes the function body chunkier and therefore more difficult to grasp.

> Also RPV is Optional so in absence of RPV  matchRPV may fail, but that does not mean anything, so the comparison has to be in correct context, hence best to be done together!

Sure, but if you look at the (pseudo)code I proposed above, this bit:
```golang
        if errors.Is(err, ErrNotFound) {
            return nil
        }
```
deals with optionality.

scheme/arm-cca/store_handler.go
thomas-fossati: I mean the caller gets back:
```
[ a, b, c ]
```
How does it tell if "a" is a PID or RID?

yogeshbdeshpande: Thanks for clarifying the question. At this point in context the GetRefValueIDs() return an array of ReferenceValueIDs for the composite attester, to fetch the Reference Values as a collection. It does not need to know the distinction, which one is Platform and which one is Realm.

thomas-fossati: OK, the caller may be just a conduit, but someone in the end must make sense of the list `[ "a", "b", "c" ]`, right?

yogeshbdeshpande: The flow is identical, whether the Attester is a Composite Attester or a Single Attester...

thomas-fossati: Cool, thanks. That answers my question.

integration-tests/data/results/cca.freshness-fail.json
yogeshbdeshpande: Done all the JSON files proper indentation!

scheme/arm-cca/evidence_handler.go
yogeshbdeshpande: @thomas-fossati : The suggested change makes the PerformAppraisal() methid look different between, `CCA Platform` & `CCA Realm` which I wanted to avoid, hence I passed the appraisal structure pointer as first argument, which will be modified by the `subAttesters`

scheme/common/cca/realm/realm_handlers.go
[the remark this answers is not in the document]
yogeshbdeshpande: Corrected in: https://github.com/veraison/services/commit/33f2ed4b6e6aeb7a20a3582a00a9445d97276291

scheme/arm-cca/store_handler.go
yogeshbdeshpande: Corrected in: https://github.com/veraison/services/commit/33f2ed4b6e6aeb7a20a3582a00a9445d97276291

scheme/arm-cca/store_handler.go
yogeshbdeshpande: Thinking a bit more I would change it to following:
Platform SW Components:
`platform.sw-component`

Platform Configuration:
`platform.config`

Realm Reference Value:

`realm.reference-value`

scheme/arm-cca/evidence_handler.go
thomas-fossati: I understand the "uniformity of interfaces" argument, but that doesn't prevent modifying both :-)

The point is that there is no reason for the caller pre-allocate the object, it's entirely the responsibility of the `PerformAppraisal` functions to do that.

scheme/common/arm/platform/iMeasurement.go
thomas-fossati, line 1: nit: file names should be lowercase (`imeasurement.go`)

thomas-fossati (commented):
sorry for the late comment, I just noticed that...

scheme/arm-cca/corim_extractor.go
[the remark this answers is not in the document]
yogeshbdeshpande: Fixed in https://github.com/veraison/services/commit/6a4041ff7758e0830e5215ba5e7ad258fa4464ea

scheme/common/cca/platform/cca_ssd_extractor.go
yogeshbdeshpande: Fixed in https://github.com/veraison/services/commit/6a4041ff7758e0830e5215ba5e7ad258fa4464ea

scheme/arm-cca/realm.go
yogeshbdeshpande: Fixed in https://github.com/veraison/services/commit/6a4041ff7758e0830e5215ba5e7ad258fa4464ea

scheme/arm-cca/realm.go
yogeshbdeshpande: If one cannot extract claims from `Evidence` surely it is a Bad Evidence, isn't it?

scheme/common/cca/realm/realm_handlers.go
yogeshbdeshpande: This debug was `extremely useful while debugging issues, I have removed almost all the un-necessary prints, but this one is particularly helpful!

yogeshbdeshpande: Need to have some tracing from Front End to Back!!! Concurrent Requests across all the pipeline!!

scheme/common/cca/realm/realm_attributes.go
yogeshbdeshpande: Drop the check!

scheme/arm-cca/realm.go
yogeshbdeshpande: Changed in https://github.com/veraison/services/commit/974a19615feee71e7c2378a8075d9617efe11513

scheme/arm-cca/realm.go
yogeshbdeshpande: Removed Bad Evidence from Realm Appraisal!

yogeshbdeshpande: As discussed, we will raise a separate issue to track `end-to-end` appraisal sequence to distinguish concurrent appraisal procedures!

yogeshbdeshpande: See https://github.com/veraison/services/issues/236

scheme/common/cca/realm/realm_handlers.go
yogeshbdeshpande: See issue #236

scheme/common/cca/realm/realm_instanceattributes.go
yogeshbdeshpande: See issue #237

scheme/arm-cca/realm.go
```diff
@@ -6,98 +6,123 @@ package arm_cca
...
+			switch matchRpv(claims, &endorsement) {
+			// Note, If an Endorser does not use RPV it indicates, one Realm per RIM, which is a match
+			case nil, ErrKeyNotFound:
+				trustVector.Executables = ear.ApprovedBootClaim
+			case ErrValuesMismatch:
+				trustVector.Executables = ear.ContraindicatedRuntimeClaim
+				continue // continue looking for other RPVs matching the same RIM
+			default:
+				// for other errors finish appraisal
+				trustVector.Executables = ear.ContraindicatedRuntimeClaim
```
yogeshbdeshpande, line 55: Verifier Malfunction!

thomas-fossati (approved):
Thanks for the herculean work :-)

PR #240, opened 2024-06-28
fix(builtin): update schemes.gen.go

setrofim:
Update schemes.gen.go to include recent schemes (parsec-cca and parsec-tpm) and StoreHandler instances.

Due to git faffing with mtime's, make cannot always detect when the file needs to be re-generated automatically.

### DIFF
--- a/builtin/schemes.gen.go
+++ b/builtin/schemes.gen.go
@@ -1,24 +1,33 @@
 package builtin
 
 import (
-	"github.com/veraison/services/plugin"
+        "github.com/veraison/services/plugin"
 
-	scheme1 "github.com/veraison/services/scheme/arm-cca"
-	scheme4 "github.com/veraison/services/scheme/psa-iot"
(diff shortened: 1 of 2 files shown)


yogeshbdeshpande (approved):
LGTM!

thomas-fossati (approved):
Q: is this something that needs to be done manually each time a scheme is added/removed?  If so, probably worth documenting it in schemes/README as well

setrofim:
> Q: is this something that needs to be done manually each time a scheme is added/removed? If so, probably worth documenting it in schemes/README as well

Well, the intention was that this done automatically via `make` by detecting changes under `scheme/`, but because `git` doesn't always preserve `mtime`s, that doesn't always work. So  yes.  (the process is excruciatingly so due a ton of static analysis being performed via Python libs, so we can't just genrate the file every time during build).

I'll add the note.

PR #249, opened 2024-07-23
fix(docker): fix case mismatch warnings

setrofim:
Latest docker started reporting warnings that the "FROM" and "as" cases don't match.


thomas-fossati (approved):
picky!

PR #254, opened 2024-08-19
Add support for MySQL/MariaDB and Postgres + fixes

setrofim:
Add support for "mysql" and "pgx" drivers (for MySQL/MariaDB and
Postgres, respectively). This required the following changes to the
"sql" kvstore backend:

- "key" column had to be renamed to "kv_key", as the former is a
  reserved word in MySQL. "vals" column has also been renamed to
  "kv_val" for consistency (also, the old name implied that the column
  contained an array of values, where as it is a single string value).
- Queries are now assembled using squirrel, which allows switching the
  placeholder symbol. This is necessary because MySQL and Postgres
  expect different syntax for placeholders.

Documentation for KVStore has been amended to reflect that only three
drivers can be enabled via configuration for the services, and the
described support for other drivers is in the context of other code
using the KVStore implementation. (Plus minor line length
re-formatting.)

Native deployment has been augmented with optional config/init scripts
for the other SQL drivers, and documentation on how to enable them.

Additionally:

- Update end-to-end input corims to the new format
- Allow -f and -s flags at the same time when creating a native deployment


kvstore/sql.go
```diff
@@ -109,18 +123,24 @@ func (o SQL) Get(key string) ([]string, error) {
...
 
-	// nolint:gosec
-	// o.TableName has been checked by isSafeTblName on init
-	q := fmt.Sprintf("SELECT DISTINCT vals FROM %s WHERE key = ?", o.TableName)
+	query := sq.Select("kv_val").Distinct().
+		From(o.TableName).
+		Where(sq.Eq{"kv_key": key}).
+		PlaceholderFormat(o.Placeholder)
+
+	query_text, args, err := query.ToSql()
```
thomas-fossati, line 131: nit: https://go.dev/doc/effective_go#mixed-caps

```suggestion
	queryText, args, err := query.ToSql()
```

kvstore/sql.go
```diff
@@ -109,18 +123,24 @@ func (o SQL) Get(key string) ([]string, error) {
...
-	rows, err := o.DB.Query(q, key)
+	rows, err := o.DB.Query(query_text, args...)
 	if err != nil {
 		return nil, err
 	}
 
 	defer rows.Close()
 
-	var vals []string
+	var kv_val []string
```
thomas-fossati, line 143: ditto

kvstore/sql.go
```diff
@@ -189,11 +211,15 @@ func (o SQL) Add(key string, val string) error {
...
-	// nolint:gosec
-	// o.TableName has been checked by isSafeTblName on init
-	q := fmt.Sprintf("INSERT INTO %s(key, vals) VALUES(?, ?)", o.TableName)
+	query := sq.Insert(o.TableName).Columns("kv_key", "kv_val").
+			Values(key, val).PlaceholderFormat(o.Placeholder)
 
-	_, err := o.DB.Exec(q, key, val)
+	query_text, args, err := query.ToSql()
+	if err != nil {
+		panic(err)
```
thomas-fossati, line 219: Any reason to `panic()`?

kvstore/sql.go
```diff
@@ -217,19 +243,29 @@ func (o SQL) Set(key string, val string) error {
...
-	// nolint:gosec
-	// o.TableName has been checked by isSafeTblName on init
-	delQ := fmt.Sprintf("DELETE FROM %s WHERE key = ?", o.TableName)
+	delQuery := sq.Delete(o.TableName).
+			Where(sq.Eq{"kv_key": key}).
+			PlaceholderFormat(o.Placeholder)
+
+	query_text, args, err := delQuery.ToSql()
+	if err != nil {
+		panic(err)
```
thomas-fossati, line 252: ditto (`panic`ing in a package)

thomas-fossati (commented):
LGTM, thanks.

A couple of nits and questions inlined.

yogeshbdeshpande (approved):
LGTM!

yogeshbdeshpande:
I do not have additional comments, I looked at the code and I had same questions as asked by Thomas so once you address them, I am fine with the change!

kvstore/sql.go
setrofim: The error only currs if the query is malfomed this something that needs to be fixed entirely in this file.

kvstore/sql.go
setrofim: same reason as above -- this is a run-time check for a compile-time error.

kvstore/sql.go
thomas-fossati: It makes sense if that's the only reason - apart from OOM, of course, which would kill the runtime anyway.

PR #259, opened 2024-08-22
chore: remove support for running services in-place

setrofim:
This is now possible via the native deployment (by deploying with symlinks), and this just adds an unnecessary maintenance burden.

Note: this basically just adds a note saying that we don't support it, and points to the native deployment. The ability to do so is still there. However, if you're exercising it, you're on your own.


thomas-fossati (approved):
LGTM

s/This no longer/This is no longer/

yogeshbdeshpande (approved):
LGTM!

PR #273, opened 2024-10-11
build: enable cgo

jraman567:
cgo allows the use of C libraries in GoLang. Some packages like sqlite3 depend on this feature.

We could allow this, barring any security issues (e.g., we can't confirm the hash of a C library before linking it).

### DIFF
--- a/mk/cmd.mk
+++ b/mk/cmd.mk
@@ -36,7 +36,7 @@ _check_scheme_loader:
 	fi
 
 $(CMD): $(SRCS) $(CMD_DEPS) _check_scheme_loader _check_version
-	go build -o $(CMD) -ldflags \
+	CGO_ENABLED=1 go build -o $(CMD) -ldflags \
 	"-X 'github.com/veraison/services/config.Version=$(VERSION_FROM_GIT)' \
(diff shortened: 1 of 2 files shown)


thomas-fossati:
> cgo allows the use of C libraries in GoLang.

Skimming [`cgo` documentation](https://pkg.go.dev/cmd/cgo):

> _"The `cgo` tool is enabled by default for native builds on systems where it is expected to work. It is disabled by default when cross-compiling as well as when the `CC` environment variable is unset and the default C compiler (typically `gcc` or `clang`) cannot be found on the system `PATH`."_

So this change would only have an effect either when cross-compiling or on pretty exotic platforms, right?

> Some packages like sqlite3 depend on this feature.

I didn't know it (or I used to know and then I forgot :-)) but [you are right](https://github.com/mattn/go-sqlite3?tab=readme-ov-file#installation).  The fact we haven't observed no build failure (at least in the dev, CI and deployed platforms we know of) is likely because of the "default enable cgo" policy of the go build environment.

It seems to me @jraman567 suggested change is correct and should be merged.

@setrofim WDYT?

setrofim (approved):
LGTM

PR #274, opened 2024-10-16
fix! Rebase to latest of CoRIM

yogeshbdeshpande:


scheme/common/cca/platform/cca_ssd_extractor.go
```diff
@@ -20,57 +20,58 @@ type CcaSsdExtractor struct {
...
 
 		if err := classAttrs.FromEnvironment(rv.Environment); err != nil {
 			return nil, fmt.Errorf("could not extract PSA class attributes: %w", err)
 		}
+		for i, m := range rv.Measurements.Values {
+			if m.Key == nil {
+				return nil, fmt.Errorf("measurement key is not set at index %d ", i)
+			}
 
-		if rv.Measurement.Key == nil {
```
jraman567, line 32: Should we keep this nil test?

I'm unsure why this test was added initially. But if we perform the m.Key.IsSet() test below when the Key is not present, the service will crash.

jraman567: Never mind, I can see the nil test below. Sorry for the noise.

jraman567 (approved):
Thanks for taking the time to do this PR, @yogeshbdeshpande ! It is a lot of changes.

I skimmed through the JSON files and examined the code. It looks good to me.

PR #283, opened 2024-12-09
fix: error handling of malformed media types

### LINKED ISSUE #282: BUG: Handling malformed media type strings in the Provisioning API handler
Handling malformed media type strings in the Provisioning API handler is buggy:
1. returns 500 (internal server error) rather than 400 (bad request)
2. clears the supplied media type from the error details

I am using veraison v0.0.2412

thomas-fossati:
Fix #282


setrofim (approved):
LGTM

Issue #294, opened 2025-01-13
feature: use different ports for HTTPS and HTTP


thomas-fossati:
Currently, the REST API endpoints use the same TCP port for both HTTP and HTTPS.

The ports in use (8080, 8888, 8088) are in the `http-alt`-ish flavour.

Since #229, HTTPS is on by default, and we should transition to ports that fall under the `https-alt` category (e.g., 8443).

PR #301, opened 2025-02-17
chore!: update default port assignments to reflect TLS

setrofim:
Default ports for services were originally selected on the assumption that the services are running over HTTP. Since commit 6b44601a, services now run over HTTPS by default. This commit updates the default port assignments to reflect the now-enabled TLS.

8080 --> 8443 for verification
8888 --> 9443 for provisioning
8088 --> 10443 for management

BREAKING CHANGE: client configurations for deployments that rely on default assignments will need to be updated.

This addresses https://github.com/veraison/services/issues/294


thomas-fossati (approved):
thanks!

yogeshbdeshpande (approved):
LGTM!

yogeshbdeshpande:
Thanks for the change. Does this needs change in any documentation we have in the `veraison/docs` repo as well?

setrofim:
> Thanks for the change. Does this needs change in any documentation we have in the veraison/docs repo as well?

Hopefully not, unless we mention the defaults in there somewhere. I'll have l look. It will definitiely require changing the defaultis in any client repos we have.

yogeshbdeshpande:
> > Thanks for the change. Does this needs change in any documentation we have in the veraison/docs repo as well?
> 
> Hopefully not, unless we mention the defaults in there somewhere. I'll have l look. It will definitiely require changing the defaultis in any client repos we have.

Thank you!

PR #306, opened 2025-03-07
fix(aws): update requirements in the README

setrofim:
Remove mention of needing a VPC with two public subnets from the prerequisites section of the REAME.md for the aws deployment.

This used to be the case, but now, the VPC (and the required subnets) will be created as part of the deployment.


thomas-fossati (approved):
❤️

yogeshbdeshpande (approved):

Issue #227, opened 2024-05-25
feature: use gh "Artifact Attestations"


thomas-fossati:
G-H [artifact attestation](https://github.blog/2024-05-02-introducing-artifact-attestations-now-in-public-beta/) is now in public beta.

It seems like something we should enable to provide relying parties high quality and granular information about the veraison instance they are accepting results from.

The produced attestation would be added (or linked by) the `.well-known/veraison` API and/or to the produced EAR in the `verifier-id` object.

setrofim:
This (partially?) addressed by https://github.com/veraison/services/pull/310.

Note that the attestation is for a specific artifact (the .deb file), and not for a Verasion version. Given this,  

- I don't know how we can include this in `.well-known` or the attestation result, as Version/VTS does not "know" how it was installed. 
- Whether it even makes sense to include this, given that for a particular verions of Veraison there might be multiple values (e.g. for deb package, for rpm package, for tarball of native deployment...).
- What the utility of including this would be, as it doesn't really tell anything to the recepient: since nothing ties a running service to the artifact that installed it. The attestation is only useful if you want to validate the associated artifact, i.e. it's usefull prior to installation, not once the service is running.

PR #310, opened 2025-03-24
feat(github): generate deb package off the monthly tag

setrofim:
Add a workflow, triggered by the pushing of the monthly tag, that will create a Debian package based on the latest state of the code line.

This workflow also creates an attestation for the specified package.

This addresses https://github.com/veraison/services/issues/227

### DIFF
--- a/.github/workflows/time-package.yml
+++ b/.github/workflows/time-package.yml
@@ -1,0 +1,66 @@
+name: Automatic package creation based on the monthly tag
+on:
+  - workflow_dispatch
+  - push:
+      tags:
+        - v0.0.**
+permissions:
+  id-token: write
+  attestations: write
+  contents: read
+jobs:
+  package:
+    name: Createa Debian package
+    runs-on: ubuntu-latest
+    env:
+      GO111MODULE: on
+      CI_PIPELINE: true
+    steps:
+      - uses: actions/setup-go@v4
+        with:
+          go-version: "1.22"
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          fetch-tags: true
+          fetch-depth: 0
+      - name: Install mockgen
+        run: |
+          go install github.com/golang/mock/mockgen@v1.7.0-rc.1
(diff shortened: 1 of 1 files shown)


.github/workflows/time-package.yml
```diff
@@ -0,0 +1,66 @@
...
+          wget https://dl.smallstep.com/cli/docs-cli-install/latest/step-cli_amd64.deb -O /tmp/step-cli_amd64.deb
+          sudo dpkg -i /tmp/step-cli_amd64.deb
+      - name: create deb
+        run: |
+          mkdir /tmp/veraison-package
+          deployments/debian/deployment.sh create-deb /tmp/veraison-package
+      - name: upload deb
+        uses: actions/upload-artifact@v4
+        with:
+          name: Veraision services Debian package
```
yogeshbdeshpande, line 61:
```suggestion
          name: Veraison services Debian package
```

yogeshbdeshpande (approved):
LGTM! - minor nit

.github/workflows/time-package.yml
```diff
@@ -0,0 +1,66 @@
+name: Automatic package creation based on the monthly tag
+on:
+  - workflow_dispatch
```
thomas-fossati, line 3: Unsure, is this going to stay?

thomas-fossati (approved):
great!

thomas-fossati:
Re: addresses #227, I think we are still missing the part where the attestation is added to the EAR / returned via the .well-known API, no?

setrofim:
> Re: addresses https://github.com/veraison/services/issues/227, I think we are still missing the part where the attestation is added to the EAR / returned via the .well-known API, no?

The attestation is not of this version of Veraison, but  for specific artifcat -- the .deb package generated. I don't see

- how we can include this in .well-know in a sensible way, or
- whether it makes sense to include this since the Version running my not have been installed from this deb even if it was build off the same commit (and therefore has the same version number).

PR #318, opened 2025-04-03
fix(aws): replace hard-coded permission boundary

setrofim:
Replace the hard-coded permission boundary ARN with the parameter that was supposed to be used.


thomas-fossati (commented):
Apologies, I didn’t see the notification.

(retrospectively) LGMT :-)

PR #335, opened 2025-09-17
Create SECURITY.md

SimonFrost-Arm:
Create default security policy file to conform to github structure for security policy.


thomas-fossati (commented):
Content-wise, this is great.

However, since this applies across all repos (modulo override), I believe this should be under the .github special repo, rather than services.

See also https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/creating-a-default-community-health-file

PR #341, opened 2025-09-23
feat: Remove dependency on machine-id

### LINKED ISSUE #124: Dependency on `machine-id`
We rely on `machine-id` to generate the session ID (see: https://github.com/veraison/services/blob/main/verification/api/handler.go#L74). `machine-id` is a "systemd-ism", and this means that Veraison cannot be easily deployed on Linux systems not running `systemd`. 

We need to either remove the dependency by figuring out some alternative way of getting a node ID (MAC address of the interface we're listening on? Generate our own node id as part of deployment?), or, at the very least, document the fact that the verification service expects `/etc/machine-id` to be present in the deployment documentation.

Sukuna0007Abhi:
Title: feat: Remove dependency on machine-id

Description:
This PR implements a flexible node ID generation system to remove the hard dependency on systemd's machine-id. The new system provides multiple fallback methods to ensure reliable node identification across different environments.

Implementation details:
1. Tries to read from a persistent node ID file (/var/lib/veraison/veraison-node-id)
2. Falls back to MAC address from the first available non-loopback interface
3. Falls back to machine-id (maintains compatibility with systemd systems)
4. Finally generates and persists a random node ID

Key improvements:
- More portable: Works on non-systemd Linux systems
- Persistent: Node ID remains stable across restarts
- Flexible: Multiple fallback methods ensure reliability
- Configurable: Node ID directory can be customized via VERAISON_NODE_ID_DIR

Added new files:
- nodeid.go: Core implementation with multiple ID generation methods
- nodeid_test.go: Comprehensive unit tests for all scenarios

Modified:
- handler.go: Updated to use the new node ID implementation

This makes the verification service more portable and removes the hard dependency on systemd's machine-id while maintaining backward compatibility.

Fixes #124 

Ready for Review sir @thomas-fossati @setrofim @yogeshbdeshpande

### DIFF
--- a/verification/api/handler.go
+++ b/verification/api/handler.go
@@ -71,12 +71,12 @@ var (
 // sequence and current time.  Routing to the correct node can therefore happen
 // based on the NodeID part of the UUID (i.e., octets 10-15).
 func mintSessionID() (uuid.UUID, error) {
-	mid, err := machineid.ID()
+	nodeID, err := getNodeID()
 	if err != nil {
-		return uuid.UUID{}, err
+		return uuid.UUID{}, fmt.Errorf("failed to get node ID: %v", err)
 	}
 
-	uuid.SetNodeID([]byte(mid))
+	uuid.SetNodeID(nodeID)
 
 	return uuid.NewUUID()
 }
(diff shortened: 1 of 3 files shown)


Sukuna0007Abhi:
Pls review in your free time sir @setrofim sir @yogeshbdeshpande sir @thomas-fossati

THS-on:
@Sukuna0007Abhi any plans on revisiting this PR?

PR #343, opened 2025-09-23
Fix TrustedServices empty reference ID handling and integration test base64 encoding

### LINKED ISSUE #42: BUG: Trusted Services errors when there are no softwareIds in the evidence context
### What version of the package are you using?
latest

### Does this issue reproduce with the latest release?
Yes

### What OS and CPU architecture are you using (`go env`)?
irrelevant

### What did you do?
Set up a TrustAnchor (using the provisioning process) with no CoSWIDs. Thus, there is not `SoftwareID`.
The system is set up for the Endorsement Store to use the memory back end.

### What did you expect to see?
Calls to `GRPC.GetAttestation` with a token for that TrustAnchor to gracefully handle the case, and not produce an error.

### What did you see instead?
When `GRPC.GetAttestation` is called with a token for that TrustAnchor, it calls `GRPC.extractEvidence`, and `extractEvidence` returns a `proto.EvidenceContext` with `SoftwareID` set to `""`.
Then, back in `GetAttestation`, the call to `o.EnStore.Get`, with `key` set to `""` returns error `"The supplied key is empty"`.

This causes `GetAttestation` to return the error.

It appears that `santizeK` is not prepared to handle the case when the `key` is empty. Possible solutions are to change that behavior in `santizeK`, or to add a check in `GetAttestation` for an empty `ec.SoftwareID` before calling `o.EnStore.Get()`.

kallal79:
Description:
This PR addresses two related issues:

## Issue #42: TrustedServices errors when no softwareIds in evidence context

**Problem**: When attestation schemes return empty reference value IDs (common when no software components are provisioned), `GetAttestation` fails with "the supplied key is empty" error from kvstore.

**Solution**: Added validation in `GetAttestation` to skip empty reference IDs before calling `kvstore.Get()`.

**Impact**: Allows attestation to succeed in environments without software components provisioned.

## Integration Test Base64 Encoding Issue

**Problem**: PSA integration tests were failing with "illegal base64 data" when `evcli psa create` processed URL-safe base64 nonces.

**Solution**: Convert URL-safe base64 nonces to standard base64 before PSA token generation, matching existing CCA implementation.

**Impact**: Fixes integration test failures for PSA scheme.

## Testing
-  All VTS tests pass
-  All PSA and CCA scheme tests pass  
-  All verification and provisioning tests pass
-  Policy tests pass
-  Both fixes tested on clean upstream main branch

## Related Issues
Fixes #42

### DIFF
--- a/vts/trustedservices/trustedservices_grpc.go
+++ b/vts/trustedservices/trustedservices_grpc.go
@@ -428,13 +428,21 @@ func (o *GRPC) GetAttestation(
 		return o.finalize(appraisal, err)
 	}
 
+	// Filter out empty reference IDs (can occur when no software components are provisioned)
+	filteredReferenceIDs := make([]string, 0, len(referenceIDs))
+	for _, refID := range referenceIDs {
+		if refID != "" {
+			filteredReferenceIDs = append(filteredReferenceIDs, refID)
+		}
+	}
+
 	appraisal.EvidenceContext.Evidence, err = structpb.NewStruct(claims)
 	if err != nil {
 		err = fmt.Errorf("unserializable claims in result: %w", err)
 		return o.finalize(appraisal, err)
 	}
 
-	appraisal.EvidenceContext.ReferenceIds = referenceIDs
+	appraisal.EvidenceContext.ReferenceIds = filteredReferenceIDs
 
 	o.logger.Debugw("constructed evidence context",
 		"software-id", appraisal.EvidenceContext.ReferenceIds,
@@ -442,11 +450,6 @@ func (o *GRPC) GetAttestation(
 
 	var multEndorsements []string
 	for _, refvalID := range appraisal.EvidenceContext.ReferenceIds {
-		// Skip empty reference IDs (can occur when no software components are provisioned)
-		if refvalID == "" {
-			o.logger.Debugw("skipping empty reference ID", "refvalID", refvalID)
-			continue
-		}
 
 		endorsements, err := o.EnStore.Get(refvalID)
 		if err != nil && !errors.Is(err, kvstore.ErrKeyNotFound) {


vts/trustedservices/trustedservices_grpc.go
```diff
@@ -442,6 +442,11 @@ func (o *GRPC) GetAttestation(
 
 	var multEndorsements []string
 	for _, refvalID := range appraisal.EvidenceContext.ReferenceIds {
+		// Skip empty reference IDs (can occur when no software components are provisioned)
```
setrofim, line 445: This does not seem right. If there are no software compoments, then surely, `appraisal.EvidenceContext.ReferenceIds` itself should be empty, rather than contain an empty string?

setrofim: Updated code still filters out empty string "IDs". My point was that I'm unclear on how they can occur -- when there no s/w components, the `ReferenceIds` list should just be empty.  The fact that empty strings can occur looks like a but, and the correct solution would be to senure they're not inserted in the first place, rather than filtered out here.

7908837174: New commit hashes:
777571e - Fix integration tests (base64 conversion) .
2f48d8c - Skip empty reference value IDs .
ac343f6 - Filter empty reference IDs at source .

setrofim: ???

Those are the commits I was refering to in my previsou message. You should NOT be checking for empty refID's when accessing 1appraisal.EvidenceContext.ReferenceIds`. Those should not be in the list to begin with. The correct fix is not to skip/filter them out, but to prevent them from being added in the first place.

vts/trustedservices/trustedservices_grpc.go
```diff
@@ -428,13 +428,21 @@ func (o *GRPC) GetAttestation(
 		return o.finalize(appraisal, err)
 	}
 
+	// Filter out empty reference IDs (can occur when no software components are provisioned)
+	filteredReferenceIDs := make([]string, 0, len(referenceIDs))
```
setrofim, line 432: This filtering shoudn't be necessary.

7908837174: fix

PR #349, opened 2025-10-02
feat: Add versioning support to plugins and scheme implementations (#120)

### LINKED ISSUE #120: Versioning for plugins/scheme implementations.
We need to add versioning to The pluggable interfaces. Versions should be logged and exposed via a meta query interces to allow clients to understand the level of support associated with particular formats.

Open question: should the version be assoicated with a scheme (i.e. all pluggables associated with a scheme--e.g. both, evidence and endorsement handlers) share a version (in the same way they share the scheme name), is each interface versioned separately?

Sukuna0007Abhi:
Plugin Versioning Implementation

Implements Issue #120 - Versioning for plugins/scheme implementations

Summary

This PR adds comprehensive versioning support to Veraison's plugin system, enabling clients to query and understand the level of support and capabilities associated with specific plugin implementations.

Requirements Addressed

- Add versioning to pluggable interfaces
- Log versions during plugin discovery  
- Expose versions via meta query interface (ServiceState API)
- Enable clients to understand support levels for particular formats

Implementation Approach

Scheme-level versioning: All pluggables associated with a scheme (evidence handler, endorsement handler, store handler) share the same version, consistent with how they share the scheme name.

Key Changes

1. Core Plugin Interface
Added GetVersion() string method to IPluggable interface. All plugins must now return a semantic version string (e.g., "1.0.0").

2. Version Query API

Manager Interface (plugin/imanager.go):
- GetPluginVersion(name string) (string, error)
- GetSchemeVersion(scheme string) (string, error)

ServiceState API (proto/state.proto):
- map<string, string> scheme_versions = 4;

3. Version Logging
Plugin versions are logged at INFO level during discovery:
loaded plugin | name=psa-evidence-handler scheme=PSA_IOT version=1.0.0 path=...

4. Scheme Versions

All attestation schemes updated to version 1.0.0:
PSA_IOT, ARM_CCA, PARSEC_CCA, PARSEC_TPM, RIOT, TPM_ENACTTRUST

Files Modified

43 files total:
- 11 core framework files (plugin system, managers, loaders, RPC channels)
- 2 proto files (definition and generated code)
- 24 scheme implementation files (6 schemes with handlers)
- 6 test infrastructure files

Usage Examples

Programmatic Query:
version, err := pluginManager.GetPluginVersion("psa-evidence-handler")
version, err := pluginManager.GetSchemeVersion("PSA_IOT")

REST API Query:
GET /verification/v1/state

Response includes scheme_versions map with all scheme versions.

Testing

All packages compile successfully with no errors or warnings. Test infrastructure has been updated to support versioning.

Breaking Changes

This is a breaking change for existing plugin implementations. All plugins must now implement the GetVersion() string method. External plugins will need to be updated accordingly.

Documentation

Added PLUGIN_VERSIONING.md with detailed implementation notes and usage examples.

Closes #120
Regards - Abhijit Das...

### DIFF
--- a/scheme/arm-cca/scheme.go
+++ b/scheme/arm-cca/scheme.go
@@ -2,7 +2,10 @@
 // SPDX-License-Identifier: Apache-2.0
 package arm_cca
 
-const SchemeName = "ARM_CCA"
+const (
+	SchemeName    = "ARM_CCA"
+	SchemeVersion = "1.0.0"
+)
 
 var (
 	EndorsementMediaTypes = []string{
--- a/PLUGIN_VERSIONING.md
+++ b/PLUGIN_VERSIONING.md
@@ -1,0 +1,104 @@
+# Plugin Versioning Implementation (Issue #120)
+
+## Summary
+
+This PR implements versioning support for Veraison plugins and scheme implementations, addressing **Issue #120**.
+
+**Issue Requirements:**
+- ✅ Add versioning to pluggable interfaces
+- ✅ Log versions during plugin discovery
+- ✅ Expose versions via meta query interface (ServiceState API)
+- ✅ Allow clients to understand level of support for particular formats
+
+## Implementation Approach
+
+**Scheme-Level Versioning:** All pluggables associated with a scheme (evidence handler, endorsement handler, store handler) share the same version, consistent with how they share the scheme name.
+
+## Key Changes
+
+### 1. Core Plugin Interface
+- Added `GetVersion() string` to `IPluggable` interface
+- All plugins must now return semantic version (e.g., "1.0.0")
+
+### 2. Version Exposure Methods
+
+**Manager Interface (`plugin/imanager.go`):**
+```go
+GetPluginVersion(name string) (string, error)
+GetSchemeVersion(scheme string) (string, error)
+```
+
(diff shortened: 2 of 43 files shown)


scheme/arm-cca/scheme.go
```diff
@@ -2,7 +2,10 @@
 // SPDX-License-Identifier: Apache-2.0
 package arm_cca
 
-const SchemeName = "ARM_CCA"
+const (
+	SchemeName    = "ARM_CCA"
+	SchemeVersion = "1.0.0"
```
setrofim, line 7: Please change all versions to `0.0.1`. (Version `1.0.0` implies that the implementation is final, and currently none of them are).

setrofim (commented):
There appears to be some confusion as to what the version pertains. The documentation/API seems to imply that both plugins and schemes have versions, however the implementation implies the version is for the scheme.

This should be made clear in the API/docstrings.

(IMO, there is no need to version plugins separately, and version ought to be associated with the scheme).

scheme/arm-cca/scheme.go
Sukuna0007Abhi: Sure sir @setrofim

Issue #179, opened 2023-07-26
Composite keys for the key-value store.


setrofim:
Currently, the K-V store uses simple string keys. Since the key acts the retrieval index, this limits us to one efficient access pattern. This ok if we only assume retrieval as part of the provisioning/verification pipelines, however makes it awkward to implement more flexible access patterns that might be needed for management operations. Essentially, if we want to use the K-V store as more than a simple cache, the current interface is insufficient.

We could expose a more database-like interface instead, allowing running arbitrary queries, however that would limit potential implementations (that would need to support the query language).

An alternative would be to allow composite keys. That is a key would be trated as a vector/tuple with each field having meaning specific to a particular application.  In addition, there would be a reserved value that can be applied to any field that means "ignore this field" when creating the access query.

Conceptually, our current K-V store looks like this:
```
key -> [value0, ..., valuen]
```
A single key maps onto zero or more values (our keys are not unique). When accessing values, we specify the `key` as the "query".

 The proposed change is to
```
[key_elt0, ...,  key_eltn] -> [value0, ..., valuen]
```
A key is comprised of multiple elements that together form the overall key under which the value is stored. When accessing values we can use the reserved "any" value in place of any of the elements as part of the "query". As a shortcut, we can also say that if the query is shorter than the total number of elements, all values for elements past the end of the query are assumed to be "any".  E.g. accessing `[v0, *, vs]`  would return all values who's keys have `key_elt0` set to `v0`, `key_elt2` set to `v2`, and `key_elt1` as well as `key_elt3` to `key_eltn` can be any value.

### Example use case: policy management

There are two major access patterns for policies -- based on tenant_id/scheme, and based on their UUID. Since the former is used during verification, it is considered more important and so our current policy store key is in the form `tenant_id:scheme`. The problem is that in order to access a policy with a specific UUID, we need to retrieved all policies for the associated tenant-id/scheme form the store, and then filter the results in code.

With composite keys, our key would be `[tenant-id, scheme, uuid]`. During verification, we can retrieve the relevant policies using something like `["myteant", "myscheme", *]`, i.e. indicating "any" uuid. For by-uuid retrieveal, the query key becomes `["myteant", *, "aaa-bb-cc-dddd"], i.e. specifying only the uuid and tenant, and indicating the scheme as "any".

### Implementation considerations

[...]

PR #357, opened 2025-10-12
feat: add & delete APIs for provisioning interface

### LINKED ISSUE #223: Feature: Implement GET & DELETE APIs for Veraison Provisioning Interface
Implement GET & DELETE APIs for Veraison Provisioning Interface!

Current Interface to Veraison Provisioning only allows Submitting Endorsements (Reference Values and Trust Anchors)!

However one should be able to Fetch these endorsements, to check, whether the submissions are as per the expectations
as well as to query what all has been submitted so far!

In addition to this an Endorser should be able to DELETE the Endorsements, once submitted.

This issue tracks this enhancements to introduce GET & DELETE APIs.

Please note: This issue was to be migrated from Veraison/veraison repository where originally it was created,
but got lost in transit!

joydeep049:
Fixes #223 

Local Tests Passing

cc @yogeshbdeshpande @thomas-fossati @setrofim


setrofim (changes requested):
Note: our current  K-V store design does not really allow for a reasonable implementation of GET/DELETE APIs. I think it may be worth  re-visiting [composite keys](https://github.com/veraison/services/issues/179) for the K-V store before tackling this (though that requires some more design work).

PR #372, opened 2026-01-26
Lead verifier: Veraison client plugin

### LINKED ISSUE #370: Lead verifier: Veraison Plugin
This issue tracks the implementation of Lead verifier's "Veraison client plugin".

The plugin must satisfy the following interface:

```golang
type IComponentVerifierClientHandler interface {
	plugin.IPluggable

	AppraiseComponentEvidence(
      evidence []byte,
      mediaType string,
      nonce []byte,
      clientCfg []byte,
  ) (*ear.Appraisal, error)
}
```

The plugin will implement a Veraison challenge-response API client in RP mode using the [`veraison/apiclient`](https://github.com/veraison/apiclient/blob/main/verification/challengeresponse.go) package.

The plugin is tasked with the following actions:

1. Receive component evidence and the nonce from the CE handler.
1. Get the verifier's public key and C-R session endpoint by querying the well-known interface.
1. Initiate a challenge-response session in RP mode with the configured verifier, supplying the component evidence and nonce.
1. Obtain an EAR from the verifier.
1. Verify the signature of the EAR.
1. Return the EAR appraisal to the CE Handler.

### Configuration

The `clientCfg` parameter supplied by the CE handler contains the relevant connectivity and trust settings as a serialised JSON byte string.

When de-serialised, the JSON object contains the following keys:

* `"url"` (mandatory): the verifier’s discovery URL
* `"ca-certs"` (optional): one or more files containing the trust anchors used to authenticate server certificates
* `"insecure"` (optional): whether certificate verification can skip the trust-related settings

Example:

```json
{
  "url": "https://downstream-verifier.example:8443/.well-known/veraison/verification",
  "ca-certs": [ "/path/to/ca1.pem", "/path/to/ca2.pem" ]
}
```

thomas-fossati:
This PR implements the “veraison client” plugin for a lead verifier.  The relevant commit is b327157.

Fix #370

Current limitation: the signature verification is hardcoded to ES256.
We could make it a configuration parameter, but ideally, we should retrieve it from the JWT/CWT header.

(Merging is currently blocked by https://github.com/veraison/apiclient/pull/37)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@
 *.so
 *.dylib
 *.bkp
+coserv/cmd/coserv-service/coserv-service
 policy/cmd/polcli/polcli
 protogen
 provisioning/cmd/provisioning-service/provisioning-service


scheme/veraison-client/veraison_client.go
```diff
@@ -4,24 +4,225 @@
...
+	crc IVeraisonChallengeResponseClient,
+) ([]byte, error) {
+	// 1. Unpack the clientCfg to obtain the verifier transport and trust
+	//    settings.
+	cfg, err := unpackConfig(clientCfg)
+	if err != nil {
+		return nil, fmt.Errorf("invalid client configuration: %w", err)
+	}
+
+	// 2. Get the verifier's public key and the C-R endpoint by querying the
```
yogeshbdeshpande: Why does each Component Evidence needs to re-fetch the key, we should have some mechanism of caching and checking for its availability, otherwise, this is a deployment bottleneck..., 

A better approach is:

1. Check for an existing key in the Cache:
2. Try for Verifying Attestation Results returned from Veraison using the key.
3. If Verification fails, Get the key, notice if changed and Verify Again..

This avoids repeated calls to Discovery Interface.

scheme/veraison-client/iveraisondiscoveryclient.go
```diff
@@ -0,0 +1,15 @@
+// Copyright 2026 Contributors to the Veraison project.
+// SPDX-License-Identifier: Apache-2.0
+
+package veraisonclient
+
+import "github.com/veraison/apiclient/verification"
+
+// IVeraisonDiscoveryClient is an interface for dealing with Veraison's
+// discovery API client
+type IVeraisonDiscoveryClient interface {
```
yogeshbdeshpande: Please add comments to each of the Interface Methods

scheme/veraison-client/iveraisonchallengeresponseclient.go
```diff
@@ -0,0 +1,19 @@
+// Copyright 2026 Contributors to the Veraison project.
+// SPDX-License-Identifier: Apache-2.0
+
+package veraisonclient
+
+import "github.com/veraison/apiclient/verification"
+
+// IVeraisonChallengeResponseClient is an interface for dealing with Veraison's
```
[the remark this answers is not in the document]
yogeshbdeshpande: Thinking a bit more, these Interfaces should be defined as part of apiclient/verification so that the Objects interacting the the apiclient should invoke this interface. Defining this here,  duplicates what is required by Challenge Response Client and then they need to be maintained at two places, leading to maintenance issues!

yogeshbdeshpande (changes requested):
Let us discuss few aspects before finalizing these code changes!

PR #396, opened 2026-03-13
refactor(vts): extract (reusable) appraisal logic into separate function

thomas-fossati:
Extract (most) of the appraisal logic from `GRPC::GetAttestation()` into `GRPC::getAttestation()` so that it can be reused from other RPC service endpoints, e.g., when appraising composite evidence.


setrofim (approved):
LGTM

yogeshbdeshpande (approved):
LGTM!

PR #408, opened 2026-04-09
Plugin initialization and configuration

setrofim:
Add an `Init()` method to `IPluggable`. This gets invoked immediately after a plugin is first loaded, giving opportunity for early one-time initialization. This method takes a `Parameters` object that contain configuration for the plugin.

`SchemeImplementationWrapper` has been updated to provide a default implementation, so implementing Init() is optional for attestation schemes.

Also, update the `lint` Makefile target to download and use a local version of `golangci-lint` to avoid possible compatibility issues with the system-installed version.

### DIFF
--- a/plugin/parameters.go
+++ b/plugin/parameters.go
@@ -1,0 +1,411 @@
+// Copyright 2026 Contributors to the Veraison project.
+// SPDX-License-Identifier: Apache-2.0
+package plugin
+
+import (
+	"encoding/base64"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"maps"
+	"math"
+
+	"github.com/spf13/viper"
+)
+
+var (
+	ErrNotSet = errors.New("parameter not set")
+	ErrInvalid = errors.New("invalid parameter value")
+)
+
+// ParametersMapFromViper constructs a map[string]*Parameters from the
+// provided *viper.Viper, where the top-level entries in the Viper become
+// map keys and the corresponding nested entries are used to construct
+// the Parameters.
+// If keyTrans function is not nil, then it is called for each top-level
+// Viper entry, and the output is used as map keys, rather than using the
+// entry directly.
+//
+// For example:
+//
+//   v := viper.New()
+//   v.Set("s1.p1", "foo")
+//   v.Set("s2.p1", 1)
+//   m1, err := ParametersMapFromViper(v, func(n string) string { return strings.ToUpper(n) })
+//
+//   m2 := map[string]*Parameters{
+//       "S1": NewParameters().SetString("p1", "foo"),
+//       "S2": NewParameters().SetInt("p1", 1),
+//   }
+//
+// m1 and m2 above are equivalent.
+func ParametersMapFromViper(v *viper.Viper, keyTrans func(string)string) (map[string]*Parameters, error) {
+	ret := make(map[string]*Parameters)
+	if v == nil {
+		return ret, nil
+	}
+
+	for key := range v.AllSettings() {
+		if sub := v.Sub(key); sub != nil {
+			params, err := ParametersFromViper(sub)
+			if err != nil {
+				return nil, fmt.Errorf("entry %q: %w", key, err)
+			}
+
+			if keyTrans != nil {
+				key = keyTrans(key)
+			}
+
+			ret[key] = params
+		} else {
+			return nil, fmt.Errorf("entry %q is not a map", key)
+		}
+	}
+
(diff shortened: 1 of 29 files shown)


plugin/parameters.go
```diff
@@ -0,0 +1,410 @@
...
+	"github.com/spf13/viper"
+)
+
+var (
+	ErrNotSet = errors.New("parameter not set")
+	ErrInvalid = errors.New("invalid parameter value")
+)
+
+// ParametersMapFromViper constructs a map[string]*Parameters from the
+// provided *viper.Viper, where the top-level entires in the Viper become
```
yogeshbdeshpande:
```suggestion
// provided *viper.Viper, where the top-level entries in the Viper become
```

plugin/parameters.go
```diff
@@ -0,0 +1,410 @@
...
+	}
+
+	o.values = tmp.values
+	return nil
+}
+
+// SetString sets the specified key to the specified string value and returns a
+// pointer to the Parameters object. If the key already exists, it is
+// overwritten.
+func (o *Parameters) SetString(key, value string) *Parameters {
```
yogeshbdeshpande, line 138: I am little confused, as to why you need functions on line 138, 145, 152, 159, 166???

Shouldn't the function Set(key string, val any) is enough to handle supported types?

yogeshbdeshpande (commented):
left some comments, will complete review by tomorrow (10/04)

plugin/parameters.go
setrofim: `Set()` may return an error if you give it an unsupported type. The typed `SetXXX()` methods cannot error, and so can return `*Paremeters`, allowing chaining, which allows for a more convenient construction of `Parameters` (e.g. see test code); or, alternatively, you can just ignore the return value since you don't have to worry about error checking.

yogeshbdeshpande: Not sure, providing alternative methods to accomplish the same objective is a great idea. The documentation on // Set() can be improved to state clearly of the supported types and indicating an erroneous results (if passed with un-supported types) thus setting the correct expectations.

setrofim: > Not sure, providing alternative methods to accomplish the same objective is a great idea. 

Why not? The typed `SetXXX()` methods are more convenient, and what you would typically want to use if you know the type of what you're putting into parameters since you don't have to do (superfluous) error checking. `Set()` is more flexible and more convenient when you don't know the type of you're putting in, since you don't have to resolve the type beforehand. 

> Set() can be improved to state clearly of the supported types and indicating an erroneous results (if passed with un-supported types) thus setting the correct expectations

Done.

jraman567 (approved):
Thanks for the PR, @setrofim ! It looks good to me.

Just to summarize, the PR addresses two of the issues we were discussing about scheme initialization/loading.
- If scheme's Init fails, this PR skips that scheme without crashing the process
- It also supplies scheme-specific parameters

I understand @yogeshbdeshpande is still reviewing, so I'd wait for his approval also. Thank you!

vts/cmd/vts-service/README.md
```diff
@@ -30,6 +31,31 @@ The following directives are currently supported:
...
+    bar: "baz"
+```
+
+Parameter values must be either strings, integers or booleans. Bytes values are
+also supported as either raw base64-encoded strings, `my_param: "3q2+7w=="`, or
+as YAML `!!binary` directives, `my_param: !!binary "3q2+7w=="`.
+
+Valid parameter names, and corresponding expected value types, are
+scheme-specific. Please refer to scheme documentation if it exists, otherwise
+assume that no parameters are supported (which is the case for most existing
```
yogeshbdeshpande:
```suggestion
assume that no parameters are provided/supplied (which is the case for most existing
```

setrofim: I don't think this is correct. What parameters provided/supplied is the function of the config file, no of the scheme implementation. Nothings stops me from providing/ supplying parameters for ARM_CCA in the config; they will just be ignored since ARM_CCA doesn't support any.

I think "supported" is correct here, but I can change to "expected" if you prefer?

yogeshbdeshpande (approved):
LGTM

vts/cmd/vts-service/README.md
yogeshbdeshpande: Sure thanks, expected sounds more reasonable but I agree, with your explaination.

thomas-fossati (approved):
Both commits LGTM, thanks!

Issue #410, opened 2026-04-16
BUG: Update TODO place holder in scheme/example


jraman567:
### What version of the package are you using?
latest source

### Does this issue reproduce with the latest release?
Yes

### What OS and CPU architecture are you using (`go env`)?
Oracle Linux 9; x86_64

### What did you do?
I tried to update go modules, and ran "go mod tidy"

### What did you expect to see?
no errors

### What did you see instead?
Following error:
github.com/veraison/services/scheme/<TODO>: malformed import path "github.com/veraison/services/scheme/<TODO>": invalid char '<'

jraman567:
@setrofim How did you want to name this package?

setrofim:
From my testing this seems to only happen for `go` versions `1.24` (and lower).  https://github.com/veraison/services/pull/411  (for unrelated reasons) updated the require version for `services` to `1.25`.

Please could you update your `go`  and confirm whether this is still an issue?

jraman567:
Thanks, @setrofim ! I can't reproduce this issue after updating go to 1.25.

Closing the issue, thank you!

PR #411, opened 2026-04-16
chore(deps): update grpc and pgx deps

cowbon:
Update dependencies to address advisories. Note pgx requires go >= 1.25.0, also bump go version as such.
[GHSA-p77j-4mvh-x3m3](https://github.com/advisories/GHSA-p77j-4mvh-x3m3)(CVE-2026-33186) 
[GHSA-xgrm-4fwx-7qm8](https://github.com/advisories/GHSA-xgrm-4fwx-7qm8)(CVE-2026-33815) 
[GHSA-9jj7-4m8r-rfcm](https://github.com/advisories/GHSA-9jj7-4m8r-rfcm)(CVE-2026-33816)

PR #412, opened 2026-04-17
fix(scheme): comment TODO import in example

setrofim:
While excluded form the build, the import causes issues when running "go mod tidy" with go version prior to 1.25.

Addresses https://github.com/veraison/services/issues/410


setrofim:
Actually, just noticed https://github.com/veraison/services/pull/411

If we're going to bump `go` requirement to `v1.25` anyway, then I don't thin this is needed.

setrofim:
No longer relevant since https://github.com/veraison/services/pull/411 advanced minimum required version to `1.25`.

Issue #414, opened 2026-05-08
BUG: Testcase failure in ARM64 architecture, but passes in x86


shubhajeet-dey-fj:
### What version of the package are you using?
Using latest `main` branch with commit id: `fb4f9fdd0094a00125081030c9b009c0b7ac863f`

### Does this issue reproduce with the latest release?
Yes, it's reproducible in arm64 arch's machines.

### System Details:
Architecture details:  `aarch64 GNU/Linux` 

`go env` output:
```
AR='ar'
CC='gcc'
CGO_CFLAGS='-O2 -g'
CGO_CPPFLAGS=''
CGO_CXXFLAGS='-O2 -g'
CGO_ENABLED='1'
CGO_FFLAGS='-O2 -g'
CGO_LDFLAGS='-O2 -g'
CXX='g++'
GCCGO='gccgo'
GO111MODULE=''
GOARCH='arm64'
GOARM64='v8.0'
GOAUTH='netrc'
GOBIN=''
GOCACHE='/home/user/.cache/go-build'
GOCACHEPROG=''
GODEBUG=''
GOENV='/home/user/.config/go/env'
GOEXE=''
GOEXPERIMENT=''
GOFIPS140='off'
GOFLAGS=''
GOGCCFLAGS='-fPIC -pthread -Wl,--no-gc-sections -fmessage-length=0 -ffile-prefix-map=/tmp/go-build1541551352=/tmp/go-build -gno-record-gcc-switches'
GOHOSTARCH='arm64'
GOHOSTOS='linux'
GOINSECURE=''
GOMOD='/home/user/workspace/services/go.mod'
GOMODCACHE='/home/user/go/pkg/mod'
GONOPROXY=''
GONOSUMDB=''
GOOS='linux'
GOPATH='/home/user/go'
GOPRIVATE=''
GOPROXY='https://proxy.golang.org,direct'
GOROOT='/usr/local/go'
GOSUMDB='sum.golang.org'
GOTELEMETRY='local'
GOTELEMETRYDIR='/home/user/.config/go/telemetry'
GOTMPDIR=''
GOTOOLCHAIN='auto'
GOTOOLDIR='/usr/local/go/pkg/tool/linux_arm64'
GOVCS=''
GOVERSION='go1.26.2'
GOWORK=''
PKG_CONFIG='pkg-config'
```
### Testcase failure
In `TestParameters_float` testcase under `/plugin/parameters_test.go` fails if the host/env is arm64 machine. 

[...]

PR #415, opened 2026-05-08
fix(plugin): fix Set() with float64

setrofim:
Correct the boundary condition for float64 value fitting into an int64 from "> float64(MaxInt64)" to ">= float64(MaxInt64)" As float64(MaxInt64) rounds to 9223372036854775808.0.

Addresses https://github.com/veraison/services/issues/414


thomas-fossati (approved):
I have checked this on arm64 (see https://github.com/veraison/services/issues/414#issuecomment-4407286402).

LGTM!

PR #416, opened 2026-05-11
Update to latest `corim-store` plus fixes

setrofim:
- Update to the latest `corim-store`, switching to it to handle CoSERV
- Fix CCA attestation handling of updated endorsement formats

### DIFF
--- a/scheme/arm-cca/scheme.go
+++ b/scheme/arm-cca/scheme.go
@@ -9,8 +9,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"strconv"
-	"strings"
 
 	"github.com/veraison/ccatoken"
 	"github.com/veraison/ccatoken/platform"
@@ -18,12 +16,15 @@ import (
 	"github.com/veraison/corim/comid"
 	"github.com/veraison/corim/profiles/cca"
 	"github.com/veraison/ear"
+	"github.com/veraison/psatoken"
 	"github.com/veraison/services/handler"
 	"github.com/veraison/services/log"
 	"github.com/veraison/services/scheme/common"
 	"github.com/veraison/services/vts/appraisal"
 	"go.uber.org/zap"
 )
+
+const NUM_REMS = 4
 
 var Descriptor = handler.SchemeDescriptor{
 	Name:         "ARM_CCA",
@@ -113,7 +114,9 @@ func (o *Implementation) GetReferenceValueIDs(
 			Class: trustAnchors[0].Environment.Class,
 		},
 		{
-			Instance: comid.MustNewBytesInstance(rimValue),
+			Class: &comid.Class{
+				ClassID: comid.MustNewBytesClassID(rimValue),
+			},
 		},
 	}, nil
 }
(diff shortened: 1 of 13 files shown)


scheme/arm-cca/scheme.go
```diff
@@ -320,57 +322,66 @@ func AppraiseRealm(
...
 				}
+				refVal.ExtensibleMeasurements[2] = digest
+			case cca.CCARealmExtendedMeasurement3Mkey:
+				digest, err := readMeasurementDigestBytes(&measurement)
+				if err != nil {
+					return fmt.Errorf("%s: %w", cca.CCARealmExtendedMeasurement3Mkey, err)
+				}
+				refVal.ExtensibleMeasurements[3] = digest
+			case cca.CCARealmPersonalizationMkey:
+				refVal.PersonalizationValue, err = measurement.Val.RawValue.GetBytes()
```
DhanusML, line 376: Shouldn't `measurement.Val.RawValue != nil` be checked before this?

scheme/arm-cca/scheme.go
DhanusML, line 567: Missing version should not be an error, since version is optional (https://www.ietf.org/archive/id/draft-ffm-rats-cca-token-03.html#name-version).

DhanusML: only compare if `len(lhsV) > 0`, since the REMs are not mandatory in realm reference values (https://www.ietf.org/archive/id/draft-ydb-rats-cca-endorsements-03.html#section-3.2.3-8).

scheme/arm-cca/scheme.go
setrofim: `GetBytes()` handles that so it's OK to call it when `RawValue` is `nil`.

DhanusML: Does it work when `Val.RawValue` is `nil`? The following code panics.

```go
package main

import (
        "encoding/json"
        "github.com/veraison/corim/comid"
)

func main() {
        js := []byte(`
{"version": {
        "value": "1.0.0",
        "scheme": 16384
 },
 "name": "foo"
}`)
        var val comid.Mval
        if err := json.Unmarshal(js, &val); err != nil {
                return
        }
        if err := val.Valid(); err != nil {
                return
        }
        val.RawValue.GetBytes()
}
```

setrofim: You're right. It only handles if the inner `val` is `nil`. My bad. I've added the check.

thomas-fossati (approved):

scheme/arm-cca/scheme.go
```diff
@@ -320,57 +323,71 @@ func AppraiseRealm(
...
+			mkey, err := readMeasurementKey(&measurement)
+			if err != nil {
+				return err
+			}
 
-			if measurement.Val.RawValue != nil {
-				refVal.PersonalizationValue, err = measurement.Val.RawValue.GetBytes()
+			switch mkey {
+			case cca.CCARealmInitialMeasurementMkey:
+				digest, err := readMeasurementDigestBytes(&measurement)
```
yogeshbdeshpande, line 341: Just wondering, we are reading via `readMeasurementDigestBytes()`,

we should also check the allowed Digest Algorithm for CCA, as ONLY three values are allowed.

cca-realm-measurement-type = bytes .size 32 / bytes .size 48 / bytes .size 64

setrofim: But this has no bearing on verification algorithm. This has already been validated by the profile-validation when the CoRIM was originally added. Here, we only care whether digests match the evidence. Structural validity of CoRIMs and evidence tokens is validated elsewhere.

yogeshbdeshpande: Sure, I wanted to check, whether when storing CoRIMs in the store this aspect has been validated or not???

scheme/arm-cca/scheme.go
yogeshbdeshpande, line 593: Orthogonal to this code review, something to remember is the case when len (lhs) can be > than len(rhs) which we should allow, but the reverse should be a failure!! Here I assume rhs is reference and lhs is Evidence

scheme/arm-cca/scheme.go
setrofim: Yes, this is validated by the profile-specific validation code.

yogeshbdeshpande (approved):
LGTM!

PR #420, opened 2026-05-27
fix(coserv): expiry + config refactor

setrofim:
- Update to v0.2.1 of corim-store that implements MaxExpiry, which is used when CoRIMs do not specify validity, and serves as a maximum limit on expiry when they do.
- In order allow configuration of MaxExpiry, refactor how configuration for CoSERV is handled inside VTS:
  - Create CoservContext which contains configuration for CoSERVService prior to its creation, and also contains the signer (which is not part of the service, but is related to it).
  - Configuration for the context is taken from "coserv" section of config files (the same as the front-end: all coserv-related configuration is now in the same place in the config tree). Configuration for the signer is moved to "signer" sub-section under it.
- FallbackAuthority for the service is now configured based on the signer's public key, rather than being hard-coded to a nonsense value.


vts/coserv/context.go
```diff
@@ -0,0 +1,121 @@
...
+
+func getMaxExpiry(v *viper.Viper) (time.Duration, error) {
+	if !v.IsSet("max-expiry") {
+		return FallbackMaxExpiry, nil
+	}
+
+	return parseDuration(v.GetString("max-expiry"))
+}
+
+func parseDuration(text string) (time.Duration, error) {
```
thomas-fossati, line 67: Why not use [`time.ParseDuration`](https://pkg.go.dev/time#ParseDuration) ?

thomas-fossati (commented):
LGTM, but I don’t get why we need to reimplement the `time.Duration` parsing logics.

vts/coserv/context.go
setrofim: Because that does not allow days, instead allowing sub-second periods that aren't relevant here.  Plus this allows for     longer indicators (e.g. `min` and `minutes`, as well as `m`), allowing for more readable configs.

thomas-fossati: OK, I guess :-)

thomas-fossati (approved):
While I am not entirely convinced that the reimplementation of the Duration parser is warranted, as the improvement in readability does not seem sufficient to justify the duplication of core library code, I can live with that.

PR #424, opened 2026-06-12
feat: add the coserv service to the native deployment

thomas-fossati:
Just as the title says.


deployments/native/README.md
```diff
@@ -286,28 +286,30 @@ local use, as they won't be configured with your host's name).
...
 ```
 
 Signing key generation relies on `step` which must be installed on the system
 (see [dependencies](#dependencies) section above).
 
 #### Using example signing key
 
-As with certificates, it is possible use the example key instead of generating
-a new one:
+As with certificates, it is possible use the example keys instead of generating
```
yogeshbdeshpande, line 308:
```suggestion
As with certificates, if possible use the example keys instead of generating
```

yogeshbdeshpande (approved):
LGTM! Minor nit

deployments/native/bin/veraison
```diff
@@ -199,6 +199,8 @@ function create_tmux_session() {
 	tmux send-keys -t $session_name:0.3 bin/management-service Space \
 	    --config Space ${_config_dir}/services/config.yaml C-m
 
+	tmux send-keys -t $session_name:0.4 bin/coserv-service Space \
+	    --config Space ${_config_dir}/services/config.yaml C-m
```
setrofim: This wouldn't work correctly as is since pane `0.4` does not exist -- commands above create a four-pane split with panes `0.0` to `0.3`, and all of them are already populated. Creating additional splits in the same tab won't be very readable, so it's better to open a new tab.

```suggestion
    tmux new-window -t $session_name -n coserv
	tmux send-keys -t $session_name:1.0 bin/coserv-service Space \
	    --config Space ${_config_dir}/services/config.yaml C-m
	tmux previous-window -t $session_name
```
(the last command command switches back to the "main" services window so it is the one seen when attaching to the session.)

deployments/native/deployment.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/bash -x
```
setrofim: Remove the `-x` (accidentally left over form debug?)

deployments/native/README.md
```diff
@@ -286,28 +286,30 @@ local use, as they won't be configured with your host's name).
...
+### Step 4: set up signing keys
 
-The signing key should be provided as `${VERAISON_ROOT}/signing/skey.jwk`. It
-will be used by the verification service to sign attestation results. The key
-must be in [JWK format](https://www.rfc-editor.org/rfc/rfc7517).
+There are two signing keys: one is used to sign the EAR containing the
+attestation results, and the other is used to sign the CoSERV query results.
+The EAR key should be provided as `${VERAISON_ROOT}/signing/skey.jwk`.  The
+CoSERV key should be provided as `${VERAISON_ROOT}/signing/coserv-signer.jwk`.
+Both keys must be in [JWK format](https://www.rfc-editor.org/rfc/rfc7517).
```
setrofim, line 295: Why use different keys for CoSERV and attestation results? They're both results produced by the service, and have the same authority.

yogeshbdeshpande: CoServ is a different service as compared to Veraison.

They have a different purpose and different role.

So it is good to keep the two separate services having two signing keys.

setrofim (approved):
Please squash the edit into the existing commit; otherwise LGTM.

deployments/native/README.md
thomas-fossati: I think this would change what is intended.
Here, we don't want to recommend using the example keys; we just want to point out that, rather than creating the credentials from scratch, one could use the pre-provided ones.
Ultimately, though, the choice is up to the user.

PR #426, opened 2026-06-16
fix(deployment/native): default corim store DB

thomas-fossati:
* store.db -> store.sql
* corim-store@… -> corim-store@v0.2.0


paulhowardarm (approved):
LGTM - many thanks!

PR #428, opened 2026-06-29
fix(verification/api): Return URL-safe nonce in challenge response sessions

cowbon:
ratsd uses URL-safe base64 encoded nonce instead of standard encoded base64. See https://github.com/veraison/ratsd/pull/37. Align the implementation so the nonce can be used directly from the `/challenge-response/v1/newSession` response body

### DIFF
--- a/integration-tests/utils/generators.py
+++ b/integration-tests/utils/generators.py
@@ -91,6 +91,10 @@ def generate_artefacts_from_response(response, scheme, evidence, signing, keys, 
     generate_expected_result_from_response(response, scheme, expected)
 
 
+def base64url_to_base64(value):
+    return value.replace('-', '+').replace('_', '/')
+
+
 def generate_expected_result_from_response(response, scheme, expected):
     os.makedirs(f'{GENDIR}/expected', exist_ok=True)
 
@@ -99,15 +103,17 @@ def generate_expected_result_from_response(response, scheme, expected):
     nonce = response.json()["nonce"]
 
     if scheme == 'psa' and nonce:
+        translated_nonce = base64url_to_base64(nonce)
         update_json(
                 infile,
-                {"PSA_IOT": {'ear.veraison.annotated-evidence': {f'psa-nonce': nonce}}},
+                {"PSA_IOT": {'ear.veraison.annotated-evidence': {f'psa-nonce': translated_nonce}}},
                 outfile,
                 )
     elif scheme == 'cca' and nonce:
+        translated_nonce = base64url_to_base64(nonce)
         update_json(
                 infile,
-                {"CCA_REALM": {'ear.veraison.annotated-evidence': {f'cca-realm-challenge': nonce}}},
+                {"CCA_REALM": {'ear.veraison.annotated-evidence': {f'cca-realm-challenge': translated_nonce}}},
                 outfile,
                 )
     else:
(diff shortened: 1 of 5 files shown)


integration-tests/utils/generators.py
```diff
@@ -138,15 +144,16 @@ def generate_evidence(scheme, evidence, nonce, signing, outname):
 
     if scheme == 'psa' and nonce:
         claims_file = f'{GENDIR}/claims/{scheme}.{evidence}.json'
+        translated_nonce = base64url_to_base64(nonce)
```
yogeshbdeshpande, line 152: @cowbon Why do we need to convert this like that.

Can we not use it directly to decode the Nonce byte 
as
base64.URLEncoding.DecodeString(encoded)

yogeshbdeshpande (commented):
please have a look!

integration-tests/utils/generators.py
cowbon: Hi Yogesh, we need Standard encoding here. The `nonce` in this function is already URL-safe encoded, so we need to decode and re-encode as standard encoding again

yogeshbdeshpande:
@cowbon : I am not clear about this change, can we have a meeting today to understand the change?

As far as I understand, Verifier does two things.

1. If it receives a Nonce Bytes, from RP it checks if it is URL Safe to Encode and then uses it, inside Session
2. If the Verifier receives nonce Size then it check for Size validity and uses to mint a URL safe Nonce value???

cowbon:
> @cowbon : I am not clear about this change, can we have a meeting today to understand the change?
> 
> As far as I understand, Verifier does two things.
> 
> 1. If it receives a Nonce Bytes, from RP it checks if it is URL Safe to Encode and then uses it, inside Session
> 2. If the Verifier receives nonce Size then it check for Size validity and uses to mint a URL safe Nonce value???

Hi Yogesh, thanks for the info about the verifier design. However, `/newSeesion` issues the standard nonce, and the burden to convert standard nonce to URL-safe nonce is on the attester. If URL-safe nonce is preferred over the standard one, why can't the verifier issue an URL-safe nonce to avoid the additional step?

verification/api/challengeresponsesession.go
```diff
@@ -69,16 +70,54 @@ type EvidenceBlob struct {
...
+	if err == nil {
+		return nonce, nil
+	}
+
+	// Keep reading sessions created before the nonce wire format switched
+	// from standard base64 to base64url.
+	if nonce, stdErr := base64.StdEncoding.DecodeString(v); stdErr == nil {
+		return nonce, nil
+	}
+
```
yogeshbdeshpande:
```suggestion

```

cowbon:
Hi @yogeshbdeshpande , thanks for sharing your thoughts. This should now reflected what we discussed.

yogeshbdeshpande:
@cowbon : As discussed let us first fix the API documentation properly before proceeding correctly with this PR!

cowbon:
> @cowbon : As discussed let us first fix the API documentation properly before proceeding correctly with this PR!

See https://github.com/veraison/docs/pull/78

integration-tests/utils/generators.py
```diff
@@ -91,6 +92,12 @@ def generate_artefacts_from_response(response, scheme, evidence, signing, keys,
     generate_expected_result_from_response(response, scheme, expected)
 
 
+def base64url_to_base64(value):
+    # evcli claim JSON uses standard base64, while session nonces are base64url.
+    decoded = base64.urlsafe_b64decode(value)
+    return base64.b64encode(decoded).decode('ascii')
```
setrofim: The only difference between the two encodings is in which non-alphanumeric characters are used for the last two ordinals. Rather then converting to bytes and back, it's faster to just substitute the specific characters in the string:
```suggestion
	replacer := strings.NewReplacer(
	    "-", "+",
	    "_", "/",
	)

	return replacer.Replace(value)
```

integration-tests/utils/generators.py
```diff
@@ -91,6 +92,16 @@ def generate_artefacts_from_response(response, scheme, evidence, signing, keys,
...
 
 
+def base64url_to_base64(value):
+    # evcli claim JSON uses standard base64, while session nonces are base64url.
+	replacer := strings.NewReplacer(
+	    "-", "+",
+	    "_", "/",
+	)
+
+	return replacer.Replace(value)
```
DhanusML: Shouldn't the padding also be added?

cowbon: Thanks for pointing that out

verification/api/challengeresponsesession.go
thomas-fossati, line 1:
```suggestion
// Copyright 2022-2026 Contributors to the Veraison project.
```

thomas-fossati (approved):
🚢 it!

Issue #427, opened 2026-06-28
`tpm-enacttrust` scheme does not enforce challenge-nonce freshness — a replayed quote still appraises as `affirming`


tyche-dev:
Hi Veraison maintainers — first, thanks for the project; the Docker deployment and the `cocli`/`evcli`/`pocli` tooling made a full provision-and-appraise loop straightforward to stand up.

While exercising the `tpm-enacttrust` scheme end-to-end (provision a CoRIM with a trust anchor + reference value, then submit a TPM quote through the challenge-response API), I noticed that the scheme appraises the **signature** and the **PCR digest versus the reference value**, but does **not** compare the nonce the attester bound into the quote's qualifying data (`TPMS_ATTEST.ExtraData`) against the session's expected nonce.

Concretely, in `scheme/tpm-enacttrust/scheme.go`, `ExtractClaims` surfaces `pcr-selection`, `hash-algorithm`, `firmware-version`, `node-id`, and `pcr-digest`, but not the quote's `ExtraData`; and `AppraiseClaims` checks only `pcr-digest` against the endorsed reference. The `eat_nonce` that appears in the EAR is the framework reflecting the session nonce, not evidence that the scheme verified freshness.

**Impact.** A correctly-signed quote over a matching PCR state, captured once, can be **replayed** to obtain `ear.status = affirming` for that instance at any later time, regardless of the instance's current state. For use cases that read `affirming` as evidence of *present* state, this is a freshness gap.

**Suggested fix (happy to send a PR).** Two small, separable pieces:
1. In `ExtractClaims`, surface the attester-bound nonce as a claim:
   ```go
   claims["nonce"] = []byte(decoded.AttestationData.ExtraData)
   ```
2. An appraisal policy (OPA/Rego) that compares that claim against the session's expected nonce (available to the policy via the session context) and returns a contraindicated verdict on mismatch.

I'd be glad to open a PR with the one-line scheme change and a policy example, and to add a `tpm-enacttrust` test vector for the replay case. Is enforcing freshness in-scheme (vs. leaving it to a policy) the direction you'd prefer? Happy to follow your steer.

Thanks!
— Anton Sokolov (Tyche Institute)

yogeshbdeshpande:
Thank you @tyche-dev for submitting the issue. We acknowledge the bug and Nonce check is essential.

You are more than Welcome to create a Pull Request and fix this bug.

tyche-dev:
[...]

PR #432, opened 2026-07-10
tpm-enacttrust: surface attester-bound nonce as a claim (#427)

flaviens:
Fixes the freshness gap reported in #427.

The `tpm-enacttrust` scheme appraises the quote signature and the PCR digest against the reference value, but never exposes the nonce the attester bound into the quote's qualifying data (`TPMS_ATTEST.ExtraData`). A correctly-signed quote captured once can therefore be replayed to obtain `affirming` at any later time.

This surfaces `ExtraData` as the `nonce` claim in `ExtractClaims`, so an appraisal policy can compare it against the session's expected nonce (the `eat_nonce` the framework reflects into the EAR) and return `contraindicated` on a stale quote. Enforcement stays in policy rather than the scheme, so the default behaviour is unchanged and operators opt in.

Note on encoding: the evidence nonce is standard base64 while the EAT `eat_nonce` is URL-safe base64, so a policy should normalise before comparing.

Includes a regression test asserting the claim is surfaced.

### DIFF
--- a/scheme/tpm-enacttrust/scheme.go
+++ b/scheme/tpm-enacttrust/scheme.go
@@ -4,6 +4,7 @@ package tpm_enacttrust
 
 import (
 	"bytes"
+	"encoding/hex"
 	"fmt"
 
 	"github.com/google/go-tpm/tpm2"
@@ -96,6 +97,17 @@ func (o *Implementation) ValidateEvidenceIntegrity(
 
 	if err = decoded.VerifySignature(pubKey); err != nil {
 		return handler.BadEvidence("could not verify evidence signature: %w", err)
+	}
+
+	// Freshness: the attester binds the session nonce into the quote's qualifying
+	// data (TPMS_ATTEST.ExtraData). A quote whose ExtraData does not match the
+	// session nonce is stale or replayed, so reject it here. See issue #427.
+	if !bytes.Equal([]byte(decoded.AttestationData.ExtraData), evidence.Nonce) {
+		return handler.BadEvidence(
+			"freshness: quote nonce (%s) does not match session nonce (%s)",
+			hex.EncodeToString(decoded.AttestationData.ExtraData),
+			hex.EncodeToString(evidence.Nonce),
+		)
 	}
 
 	return nil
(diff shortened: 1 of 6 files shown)


scheme/tpm-enacttrust/scheme.go
```diff
@@ -70,6 +70,10 @@ func (o *Implementation) ExtractClaims(
 	claims["firmware-version"] = decoded.AttestationData.FirmwareVersion
 	claims["node-id"] = decoded.NodeId.String()
 	claims["pcr-digest"] = []byte(decoded.AttestationData.AttestedQuoteInfo.PCRDigest)
+	// Surface the attester-bound qualifying data (TPMS_ATTEST.ExtraData) so an
```
yogeshbdeshpande: This is not how the Challenge is verified.

Please refer to other schemes, where the Challenge (nonce) is verified during ValidateEvidenceIntegrity() check

yogeshbdeshpande: For example see: https://github.com/veraison/services/blob/main/scheme/psa-iot/scheme.go#L142

yogeshbdeshpande (changes requested):
Please see the comments!

yogeshbdeshpande (approved):
Thanks for making the changes, LGTM!

yogeshbdeshpande:
@flaviens Please check the Integration tests, they also need modification, perhaps!

yogeshbdeshpande:
@thomas-fossati , could you please review so that we can merge!!

thomas-fossati (approved):
LGTM, thanks!

yogeshbdeshpande:
@flaviens, we approved and merged this PR, seems due to latest Nonce changes, this PR seems to have broken the Integration Tests. Can you please spare few cycles and create a new PR to fix the Integration tests....

flaviens:
Hi @yogeshbdeshpande it's been super long and I have limited memories of the context / knowledge of the repo (and limited time rn to do something high quality in a repo I'm not warm with ^^'). My fix would be AI-heavy. Lmk if you prefer do something yourselves.

flaviens:
Attempted a fix. Thanks!

PR #434, opened 2026-07-28
Add NVIDIA remote attestation scheme

jraman567:
This scheme adds verification code for NVIDIA attestation reports. I have verified it using NRAS (verifier_mode: "remote") and Trust Outpost (verifier_mode: "local").

There are a couple of issues:
- Build system needs to have NVAT library installed (as "nvat.h" is required). Is this OK, and could we updated our CI build image? Or should I make the build of this scheme conditional on availability of NVAT
- The scheme advertises RATSd's broader evidence format as supported media type. We should probably narrow it down with a profile.


scheme/nvidia/scheme.go
```diff
@@ -0,0 +1,276 @@
...
+
+	return claims, nil
+}
+
+func (o *Implementation) ValidateEvidenceIntegrity(
+	evidence *appraisal.Evidence,
+	trustAnchors []*comid.KeyTriple,
+	endorsements []*comid.ValueTriple,
+) error {
+	return nil
```
thomas-fossati, line 157: hmm 😅

scheme/nvidia/scheme.go
```diff
@@ -0,0 +1,276 @@
...
+			claims[attestationReportClaimName]))
+	}
+
+	nonce, ok := claims[nonceClaimName].([]byte)
+	if !ok {
+		return nil, handler.BadEvidence(fmt.Errorf("nonce has unexpected type %T",
+			claims[nonceClaimName]))
+	}
+
+	if o.nvat != nil {
```
thomas-fossati, line 182: Is it possible for `o.nvat` to be nil?

scheme/nvidia/nvat_gpu_interface.go
thomas-fossati, line 1: It looks like there is no explicit Close/Destroy function on the NvatGpuInterface.
How are we ensuring that the C resources are not leaked?

scheme/nvidia/nvat_bindings.go
```diff
@@ -0,0 +1,246 @@
+package nvidia
+
+/*
+#cgo LDFLAGS: -ldl
+#include <dlfcn.h>
+#include "nvat_bindings.h"
+#include <string.h>
+
+static void* nvatHandle = NULL;
+static char nvatLastError[256];
```
thomas-fossati, line 12: This is not supposed to run concurrently, right?
(Otherwise we may have a sync problem with these unprotected globals.)

scheme/nvidia/nvat_gpu_interface_test.go
```diff
@@ -0,0 +1,66 @@
...
+		t.Fatalf("failed to initialize NVAT interface: %v", err)
+	}
+
+	evidencePath := filepath.Join("testdata", "blackwell-evidence.json")
+	evidenceBytes, err := os.ReadFile(evidencePath)
+	if err != nil {
+		t.Fatalf("failed to read evidence test data: %v", err)
+	}
+
+	err = nvat.VerifyEvidence("0xcb86922a772181d54971ccc584bbc96b717c32b7087fff9aaf52832dd3f697a1", evidenceBytes)
```
thomas-fossati, line 64: why the 0x-prefix?

thomas-fossati (commented):
@jraman567 thanks for the contribution!

I have left my usual handful of comments inline.

scheme/nvidia/scheme.go
jraman567: Hi @thomas-fossati , sorry for the delay.

NVIDIA doesn't allow us to divide the verification into Integrity verification (ValidateEvidenceIntegrity) and Claims verification (AppraiseClaims). Both these operations are performed simultaneously when we call "o.nvat.VerifyEvidence".

scheme/nvidia/scheme.go
jraman567: It's not possible for `o.nvat` to be nil. The `Init` function of the implementation initializes this variable. If initialization fails, the plugin won't load.

scheme/nvidia/nvat_gpu_interface.go
jraman567: That's a good point, we don't explicitly destroy the NVAT object.

However, is that scenario likely to happen? Once the plugin's Implementation is loaded, the Implementation remains opened until the server is stopped. Or is it possible to unload the Implementation while the server is still running?

When the server is stopped, the resource should be free'd.

I'd like to add this explicit calls to destroy the NVAT object. Do you have a recommendation on where to place it?

scheme/nvidia/nvat_gpu_interface_test.go
jraman567: The underlying C API accepts a nonce with the "0x" prefix.

https://docs.nvidia.com/attestation/nv-attestation-sdk-cpp/latest/api/function_nvat_8h_1a8e6d1552b8ce49db3188328700f4c7ed.html

scheme/nvidia/nvat_bindings.go
setrofim, line 1: For consistency this file should probably be named `nvat_bindings.go` rather than `nvatBindings.go`; similarly, for `nvatGpuInteface.go` and `nvatOptions.go` below.

go.mod
```diff
@@ -47,15 +47,20 @@ require (
...
 	go.uber.org/zap v1.27.0
-	golang.org/x/text v0.37.0
+	golang.org/x/text v0.39.0
 	google.golang.org/grpc v1.79.3
 	google.golang.org/grpc/cmd/protoc-gen-go-grpc v1.1.0
 	google.golang.org/protobuf v1.36.10
 	gopkg.in/go-jose/go-jose.v2 v2.6.3
 )
 
+require (
```
setrofim: merge into existing indirect `require` block below.

setrofim (approved):

jraman567:
Hi @setrofim , thanks for the review and approving the PR. I have made the changes you recommended.

jraman567:
Hi @thomas-fossati, please let me know if this PR looks good. Thank you!

scheme/nvidia/nvat_bindings.go
jraman567: That's a good catch. In this implementation, this is only used during Init. So, it won't be called concurrently.

thomas-fossati (approved):
🚢 it!

jraman567:
Thanks, @thomas-fossati

deployments/docker/src/builder.docker
```diff
@@ -25,6 +33,12 @@ RUN apt-get update \
         nmap \
         sudo \
 	ca-certificates \
+        wget \
+    && distro="$(. /etc/os-release && printf '%s%s' "$ID" "$(printf '%s' "$VERSION_ID" | tr -d .)")" \
+    && wget --quiet "https://developer.download.nvidia.com/compute/cuda/repos/${distro}/x86_64/cuda-keyring_1.1-1_all.deb" \
```
anta5010, line 38: This change breaks Veraison docker deployments on arrm64 machines

Issue #431, opened 2026-07-09
feature: endorsement store plugin


DhanusML:
In the current implementation, VTS service owns an instance of corim-store,
which is used as the endorsement store.

The proposal is to make the store a go-plugin, similar to the scheme plugins.
The store plugins implement an _endorsement store_ interface, that includes
methods for

* adding CoRIM data to the store
* fetching CoMID triples using CoMID environment map
* responding to CoSERV queries

Store plugins are wrappers around a store backend.

```
+---------+            +---------------------+             +---------------+             +---------------+
|         |            |                     |             |               |             |               |
|   vts   |<---------->|  endorsement store  |<----------->| store plugin  |<----------->| store backend |
|         |            |        API          |             |               |             |               |
+---------+            +---------------------+             +---------------+             +---------------+
```

Possible initial implementations of store plugin are:

1. corim-store based: in this implementation, the plugin would be a wrapper around corim-store's
store API.

2. CoSERV based (#387):
in this implementation, the backend used by the plugin is an api client
for a CoSERV service. (It is also possible to convert the existing CoSERV proxy handler plugins to
store plugins)

VTS would maintain a manager for store plugins, which will select the correct store(s) to handle a
certain request. There would be a collection of stores that are available for a scheme. The
manager can select the correct store plugin(s) for that scheme using the scheme name or media types
as the identifier.

### Implementation details

The following implementations are required for a completing this feature.

1. Definition of the endorsement store interface and implementing RPC layer for it: this
includes defining the RPC channels used by the plugin.

2. Changes in VTS RPC server:  VTS should own a store plugin manager in addition to scheme plugin
manager and for each store operation load the appropriate store plugin. Loading of plugin could
be using media type or a store label, constructed using attestation scheme name.

3. Implementation of the plugins and discovery by VTS: this requires changes in the build scripts
to register the new plugins and also in the VTS service entry point to load and initialize the new
plugins.

4. Plugin management changes: in the current implementation, only one plugin can be fetched at a time
using an identifier. For the store plugins, the intended use cases might want multiple store
plugins for a scheme. Hence it should be made possible to fetch multiple schemes using an
identifier.

PR #435, opened 2026-08-04
Endorsement Store Interface Implementation

shefali-kamal:
This PR implements following:

- contains the implementation of endorsement store plugin interface
- converted coserv proxy plugins to endorsement store plugins
- make changes in vts to use the endorsement store plugins
- vts now uses a composite store made up of corimstore and store plugins as fallback

Address issue https://github.com/veraison/services/issues/431


endorsementstore/corimstore/corimstore_enstore.go
```diff
@@ -0,0 +1,140 @@
...
+
+// implement the IEndorsementStore interface for corimstore.Store
+type DefaultStore struct {
+	Store  *corimstore.Store
+	logger *zap.SugaredLogger
+}
+
+func NewStore() *DefaultStore {
+	logger := log.Named(PluginName)
+	logger.Debug("initializing default store")
```
setrofim: This log line belongs inside `DefaultStore.Init()` below, not here.

endorsementstore/corimstore/corimstore_enstore.go
```diff
@@ -0,0 +1,140 @@
...
+}
+
+func (s *DefaultStore) Close() error {
+	s.logger.Info("closing corimstore")
+	return s.Store.Close()
+}
+
+func (s *DefaultStore) Init(params *plugin.Parameters) error {
+	if params == nil {
+		panic("parameters are required for corimstore")
```
setrofim: Return an error rather than panic here.

endorsementstore/corimstore/corimstore_enstore.go
```diff
@@ -0,0 +1,140 @@
...
+	loader := config.NewLoader(&cfg)
+	if err := loader.LoadFromMap(params.Map()); err != nil {
+		return err
+	}
+
+	s.logger.Debugf("connecting to %s store %s", cfg.DBMS, cfg.DSN)
+
+	store, err := corimstore.Open(context.Background(), cfg.StoreConfig())
+	if err != nil {
+		panic(err)
```
setrofim: Return the error rather than panic.

endorsementstore/corimstore/corimstore_enstore.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+
+	// The store must be initialized before it may be used. In general, we
+	// rely on the store being pointed to by DSN to be initialized prior
+	// to starting the VTS. For in-memory store this can never be the case, so we
+	// initialize it here.
+	if strings.Contains(cfg.DSN, ":memory:") {
+		if err := store.Init(); err != nil {
+			panic(err)
```
setrofim: Return the error rather than panic.

endorsementstore/corimstore/corimstore_enstore.go
```diff
@@ -0,0 +1,140 @@
...
+
+func (s *DefaultStore) GetName() string {
+	return PluginName
+}
+
+func (s *DefaultStore) GetAttestationScheme() string {
+	return SchemeName
+}
+
+func (s *DefaultStore) GetSupportedMediaTypes() map[string][]string {
```
setrofim, line 137: I don't think the concept of a "supported media type" really makes sense for a store backend. I think we may want a different plugin interface for this.

endorsementstore/corimstore/corimstore_enstore.go
```diff
@@ -0,0 +1,140 @@
...
+
+func (s *DefaultStore) GetValueTriples(env *comid.Environment, label string, exact bool) ([]*comid.ValueTriple, error) {
+	res, err := s.Store.GetActiveValueTriples(env, label, exact)
+	if err == corimstore.ErrNoMatch {
+		return nil, handler.ErrNotFound
+	}
+	return res, err
+}
+
+func (s *DefaultStore) ExecuteCoservQuery(mediaType, query string) (*coserv.Coserv, error) {
```
setrofim, line 53: Why does this take a `mediaType`?  It does not appear to be used, and also clashes with remaining store API that takea `label` instead. The media type should be resolved to a label by the time the request reaches the store.

endorsementstore/nvidia-coserv/coserv_handler.go
```diff
@@ -184,7 +185,7 @@ func (s CoservProxyHandler) addReferenceValuesForClass(query *coserv.Query, c *c
 	return nil
 }
 
-func (s CoservProxyHandler) GetEndorsements(tenantID string, query string) ([]byte, error) {
+func (s CoservProxyHandler) ExecuteCoservQuery(mediaType, query string) (*coserv.Coserv, error) {
```
setrofim: Ditto here. the media type should be resolved by this point.

setrofim (changes requested):
I don't think the notion of a "media type" makes sense in the context of the store backend. We don't really want to be selecting the backend based on it. I also don't like the somewhat arbitrary distinction between "primary" and "fallback" backends.

Suggestion: store backends are identified solely by name; for the sake of reusing existing plugin loader/manager the media type APIs implemented via a shim to to just return the name, and this is hidden as much as possible. Store configuration contains a list of backend names,  which specifies which frontend plugins will actually be actively used by the store frontend. When servicing requests, the frontend tries the backends in the order specified until.

vts/trustedservices/trustedservices_grpc.go
```diff
@@ -175,12 +171,13 @@ func (o *GRPC) Close() error {
 		o.logger.Errorf("scheme plugin manager shutdown failed: %v", err)
 	}
 
-	if err := o.CoservProxyPluginManager.Close(); err != nil {
-		o.logger.Errorf("coserv plugin manager shutdown failed: %v", err)
+	// FIXME: close stores while closing store manager
```
yogeshbdeshpande: Why do we need Line 174..?

Should not the Sore manager be responsible for closing the Stores!

vts/trustedservices/trustedservices_grpc.go
```diff
@@ -553,7 +582,7 @@ func (c *GRPC) GetSupportedCoservMediaTypes(context.Context, *emptypb.Empty) (*p
 	mediaTypes = append(mediaTypes, corimDerived...)
 
 	coservProxyDerived := c.assembleCoservMediaTypes(
-		c.CoservProxyPluginManager.GetRegisteredMediaTypes(),
+		c.StoreManager.GetRegisteredMediaTypes(),
```
yogeshbdeshpande, line 582: Agree with Sergei, the Store should not have a Media Type. 

The functionality must be changed!

yogeshbdeshpande (commented):
Have started reviewing it, will complete by the end of the day today!

vts/trustedservices/trustedservices_grpc.go
```diff
@@ -473,37 +491,48 @@ func (o *GRPC) getKeyTriples(
...
+
 	var valueTriples []*comid.ValueTriple //nolint
 
 	for _, valID := range referenceValueIDs {
-		triples, err := o.Store.GetActiveValueTriples(valID, label, exact)
-		if err != nil && !errors.Is(err, corimstore.ErrNoMatch) {
-			return nil, err
+		triples, err := store.GetValueTriples(valID, label, exact)
+		if err != nil {
+			o.logger.Warnw("could not find in store", "valID", valID, "error", err)
```
thomas-fossati: When joining the dots between L526-527 and L532-533, it looks like _any_ store errors will be seen as ENOTFOUND from the pov of the caller.  If that's the case, the caller (L446) will log and move on.  So, imagine the case where the DB is down, valueTriples stays empty, getValueTriples returns ENOTFONUD, the caller logs a warning and proceeds with nil endorsements into AppraiseEvidence.  I don't think it's what we want, right?

vts/trustedservices/trustedservices_grpc.go
```diff
@@ -175,12 +171,13 @@ func (o *GRPC) Close() error {
...
-		o.logger.Errorf("coserv plugin manager shutdown failed: %v", err)
+	// FIXME: close stores while closing store manager
+	if err := CloseCorimStore(o.StoreManager); err != nil {
+		o.logger.Errorf("failed to close corim store: %v", err)
 	}
 
-	if err := o.Store.Close(); err != nil {
-		o.logger.Errorf("store closure failed: %v", err)
+	if err := o.StoreManager.Close(); err != nil {
+		o.logger.Errorf("store plugin manager shutdown failed: %v", err)
```
thomas-fossati: Potential ENOCOFFEE :-)  Would this effectively result in a double close?

vts/trustedservices/trustedservices_grpc.go
```diff
@@ -624,29 +653,23 @@ func getEndorsementsError(err error) *proto.EndorsementQueryOut {
...
-	coservService := corimstore.NewCoSERVService(
-		o.Store,
-		o.CoservContext.FallbackAuthority,
-		o.CoservContext.MaxExpiry,
-	)
-	if err := coservService.UpdateCoSERV(&query); err != nil {
-		return nil, err
+	resp, err := store.ExecuteCoservQuery(queryIn.MediaType, queryIn.Query)
+	if err != nil {
+		o.logger.Infof("could not find coserv result in store")
```
thomas-fossati, line 670: What is the reasoning behind masking any error as ENOTFOUND/404?
From a ReST perspective, we want to be able to tell our clinets that there is something wrong with the service (5xx) rather than with their query (4xx).

endorsementstore/corimstore/corimstore_enstore.go
```diff
@@ -0,0 +1,140 @@
...
+	s.logger.Infof("got coserv query: %v", query)
+	fallbackAuthority, err := comid.NewCryptoKeyTaggedBytes([]byte("dummyauth"))
```
thomas-fossati: curious: why we need to use this "dummyauth"?

proto/endorsement_store.proto
```diff
@@ -0,0 +1,20 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+syntax = "proto3";
+
+package proto;
+
+option go_package = "github.com/veraison/services/proto";
+
+message GetEndorsementsArgs {
+  bytes environment = 1 [json_name = "environemnt-map"];
```
thomas-fossati:
```suggestion
  bytes environment = 1 [json_name = "environment-map"];
```

vts/cmd/vts-service/main.go
```diff
@@ -42,10 +43,15 @@ func main() {
...
 
-	log.Info("initializing stores")
-	enStore, err := store.New(subs["store"], log.Named("store"))
-	if err != nil {
-		log.Fatalf("endorsement store initialization failed: %v", err)
+	var coservContext *coserv.Context
+	if subs["coserv"].IsSet("signer") {
+		coservContext, err = coserv.NewCoservContextFromViper(subs["coserv"])
+		if err != nil {
+			log.Fatal("CoSERV config initialization: %v", err)
```
thomas-fossati:
```suggestion
			log.Fatalf("CoSERV config initialization: %v", err)
```

thomas-fossati (commented):
Thanks for this!

I have inlined a few comments.

vts/trustedservices/trustedservices_grpc.go
DhanusML: That's correct. But with the current implementation of the store manager, it is not possible. This is because the pluggable interface lacks a `Close` method, which can be called at the end (similar to the `Init` method that gets called at the start). The `StoreManager.Close` method calls the [`plugin.Client.Kill`] method on the store plugin client, which terminates the plugin process, without closing the file descriptors and other resources that are in use by the plugin process. The note here was for myself and this should be fixed before the pull request can be merged.

[`plugin.Client.Kill`]: https://pkg.go.dev/github.com/hashicorp/go-plugin#Client.Kill

vts/trustedservices/trustedservices_grpc.go
DhanusML: No, `StoreManager.Close` does not properly close the corim-store (see https://github.com/veraison/services/pull/435#discussion_r3783627185). This is a hacky fix and should be fixed before merging the pr.

endorsementstore/corimstore/corimstore_enstore.go
DhanusML: This is a copy-paste of an older version of the code (before https://github.com/veraison/services/pull/420). I should update this with the latest changes. Thanks for pointing it out.

vts/trustedservices/trustedservices_grpc.go
DhanusML: The API layer seems to treats all errors from VTS as Internal Server Error.

coserv:
https://github.com/veraison/services/blob/f997c5275c3039caf1976139c726ceffe7d273b5/coserv/api/handler.go#L225-L231

verification:
https://github.com/veraison/services/blob/f997c5275c3039caf1976139c726ceffe7d273b5/verification/api/handler.go#L414-L430

But more context can be added to the error. I will try to do that.

vts/trustedservices/trustedservices_grpc.go
DhanusML: All errors other than ENOTFOUND from all the stores are logged as warning (/vts/trustedservices/store.go:122). The reason was to avoid the operation from failing some of the stores are down. But now I realized that this is a bad idea. Will update to return on unexpected store errors.

endorsementstore/corimstore/corimstore_enstore.go
DhanusML: Since store plugins can also serve CoSERV queries, the supported media types can be the CoSERV media types they support.

DhanusML:
Major changes:
* VTS config now includes an `active-stores` section containing a list of store plugins that will be used (see `/vts/endorsementstore` for details)
* Added `Fini` method in `IPluggable`

vts/cmd/vts-service/main.go
```diff
@@ -61,16 +68,32 @@ func main() {
...
+
+	if err != nil {
+		log.Fatalf("could not create endorsement store configuration: %v", err)
+	}
+
+	// for builtin loader
+	maps.Copy(pluginConfig, endorsementStoreConfig)
+
+	log.Debug("loading scheme and endorsement store plugins")
+	psubs, err := config.GetSubs(subs["plugin"], "*go-plugin", "*builtin", "go-plugin-stores")
```
thomas-fossati, line 96: Should this be made conditional instead?

I suspect this breaks when the services are built with `make SCHEME_LOADER=builtin`

endorsementstore/corimstore/corimstore_enstore.go
```diff
@@ -0,0 +1,145 @@
...
+	s.logger.Debugf("adding CoRIM")
+	return s.Store.AddBytes(data, label, activate)
+}
+
+func (s *DefaultStore) Fini() error {
+	s.logger.Info("closing corimstore")
+	if err := s.Store.Close(); err != nil {
+		s.logger.Errorf("Failed to close corim-store: %v", err)
+		return err
+	}
```
thomas-fossati, line 92: This could panic if Init failed before assigning s.Store:

```suggestion
    if s.Store != nil {
	    if err := s.Store.Close(); err != nil {
		    s.logger.Errorf("Failed to close corim-store: %v", err)
		    return err
	    }
	}
```

vts/trustedservices/trustedservices_grpc.go
thomas-fossati: Checking: do you still plan to address this?

vts/trustedservices/trustedservices_grpc.go
thomas-fossati: OK, thanks for checking.
I think that's a problem with the frontends that we'll need to fix.
However, first they need to get the right signal from VTS :-)

vts/endorsementstore/store.go
```diff
@@ -0,0 +1,174 @@
...
+		return nil, ErrNoStores
+	}
+	return s.stores, nil
+}
+
+func CreateEndorsementStore(plugins []string, manager StoreManager, logger *zap.SugaredLogger) (handler.IEndorsementStore, error) {
+	store := VtsEndorsementStore{
+		logger: logger,
+	}
+	for _, pl := range plugins {
```
thomas-fossati, line 48: I think we need to explicitly check that the plugins list is not empty here.

handler/endorsementstore_rpc.go
```diff
@@ -0,0 +1,267 @@
...
+
+func (c *StoreRPCClient) AddCorimBytes(data []byte, label string, activate bool) error {
+	args := proto.SubmitEndorsementsArgs{
+		Endorsement: data,
+		Label:       label,
+		Activate:    activate,
+	}
+	var unused any
+	if err := c.client.Call("Plugin.AddCorimBytes", &args, &unused); err != nil {
+		return err
```
thomas-fossati:
```suggestion
		return ParseError(err)
```

thomas-fossati (commented):
Thanks!  I have re-reviewed and added a few comments.

vts/trustedservices/trustedservices_grpc.go
DhanusML: The error is propagated till vts/endorsementstore/store.go, forgot to update the final layer. Will update to throw an error on any store errors other than ENOTFOUND.

vts/cmd/vts-service/main.go
DhanusML: config for endorsement store plugins are copied to `pluginConfig` in line 93.

deployments/native/README.md
```diff
@@ -365,6 +365,7 @@ structure
 ├── bin
 ├── certs
 ├── config
+├── endorsementstore
```
setrofim, line 368: I think the name should make it clearer that the directory contains plugins (rather than the actual store database). Also rather than increasing the number of top-level directories, it would good to move this under `plugin/` so that  it looks something like

```
├── plugins
    ├── schemes
    │    ├── arm-cca.plugin
    │    └── ...
    └── stores
         ├── corim-store.plugin
         └── ...
```

endorsementstore/amd-kds-coserv/coserv_handler.go
```diff
@@ -73,14 +84,10 @@ func (s CoservProxyHandler) addTrustAnchorForInstance(i *coserv.StatefulInstance
 		return err
 	}
 
-	// TODO(paulhowardarm) - This authority is a dummy value.
+	// TODO(paulhowardarm)
```
setrofim, line 87: At this point, probably remove the TODO as well

endorsementstore/amd-kds-coserv/Makefile
```diff
@@ -0,0 +1,15 @@
+# Copyright 2025-2026 Contributors to the Veraison project.
+# SPDX-License-Identifier: Apache-2.0
+
+.DEFAULT_GOAL := test
+
+GOPKG := github.com/veraison/services/endorsementstore/amd-kds-coserv
```
setrofim, line 6: Similar to directory naming in the deployment, I would change the package name to make it clearer that this is a plugin maybe `store-plugin` instead of `endorsementstore`?

endorsementstore/corimstore/plugin/main.go
setrofim, line 1: Minor nit, but for consistency call the package `corim-store` rather than `corimstore`.

endorsementstore/corimstore/corimstore_enstore.go
```diff
@@ -0,0 +1,145 @@
...
+	"go.uber.org/zap"
+)
+
+const (
+	PluginName = "corim-store"
+	SchemeName = "ANY"
+)
+
+// implement the IEndorsementStore interface for corimstore.Store
+type DefaultStore struct {
```
setrofim, line 26: Change the name to just `Store` since this isn't really default  in the context of this module.

endorsementstore/corimstore/corimstore_enstore.go
setrofim: Generally, `Fini()` should only be called if `Init()` succeeds, so it's reasonable to assume that `Store` has been initialized inside `Fini()`.

endorsementstore/corimstore/corimstore_enstore.go
setrofim: That's not really clear from the API though, and even with the old CoSERV implementations this was a bit of a hack. Now that we're officially adding a new type of plugins, unrelated to attestation schemes, I think the general plugin API needs to be rethought, rather than piggybacking on the exiting `IPluggable`  definition that was created with attestation schemes in mind.

endorsementstore/nvidia-coserv/coserv_handler.go
```diff
@@ -161,14 +172,11 @@ func (s CoservProxyHandler) addReferenceValuesForClass(query *coserv.Query, c *c
 
 				// We'll just look at reference value triples in the CoMID
 				for _, triple := range c.Triples.ReferenceValues.Values {
+					authority := s.FallbackAuthority
 					// Turn each triple into a quad
-					// TODO(paulhowardarm) - This authority is a dummy value.
+					// TODO(paulhowardarm)
```
setrofim, line 177: The TODO itself should be removed as well.

endorsementstore/corimstore/corimstore_enstore.go
thomas-fossati: OK, then let's panic() on a nil s.Store if that's an invariant

plugin/goplugin_loader.go
```diff
@@ -174,7 +174,8 @@ func DiscoverGoPluginUsing[I IPluggable](o *GoPluginLoader) error {
 
 		o.logger.Debugw("initializing plugin", "plugin", pluginName, "params", params.Map())
 		if err := pluginContext.Handle.Init(params); err != nil {
-			o.logger.Errorf("plugin q: %s", pluginName, err.Error())
+			o.logger.Errorf("plugin %q: %s", pluginName, err.Error())
+			pluginContext.Close()
```
DhanusML, line 178: `Close` calls `Fini` and then `Client.Kill`. But as per https://github.com/veraison/services/pull/435/changes#r3923652647, `Fini` should only be called during graceful exit. So, this should be changed to `pluginContext.client.Kill()`.

endorsementstore/corimstore/corimstore_enstore.go
DhanusML: Actually in the current implementation, `Fini` is also called if `Init` fails (`Close` is called to terminate the plugin server if `Init` fails ([goplugin_loader.go:177)](https://github.com/veraison/services/pull/435/changes#r3924991888)).

I will make changes there and also `panic` on a nil `s.Store` here

endorsementstore/amd-kds-coserv/coserv_handler.go
DhanusML: Only the dummy authority issue is resolved. Rest of the points are still todo

store-plugin/corim-store/corimstore_enstore.go
```diff
@@ -0,0 +1,148 @@
+// Copyright 2025-2026 Contributors to the Veraison project.
+// SPDX-License-Identifier: Apache-2.0
+package corimstore
```
yogeshbdeshpande, line 3: the package name should be `corim-store`

store-plugin/corim-store/corimstore_enstore.go
```diff
@@ -0,0 +1,148 @@
...
+	handler "github.com/veraison/services/handler"
+	"github.com/veraison/services/log"
+	"github.com/veraison/services/plugin"
+	vtsstore "github.com/veraison/services/vts/endorsementstore"
+	"go.uber.org/zap"
+)
+
+const (
+	PluginName = "corim-store"
+	SchemeName = "ANY"
```
yogeshbdeshpande, line 22: Is it mandatory to have the schemename in the plugin, in theory this is not needed..?

store-plugin/corim-store/corimstore_enstore.go
```diff
@@ -0,0 +1,148 @@
...
+	s.CoservCfg = cfg.StoreCommonParams
+
+	s.logger.Debugf("connecting to %s store %s", cfg.DBMS, cfg.DSN)
+
+	store, err := corimstore.Open(context.Background(), cfg.StoreConfig())
+	if err != nil {
+		return err
+	}
+
+	// The store must be initialized before it may be used. In general, we
```
yogeshbdeshpande, line 118:
```suggestion
	// The store must be initialized before it can be used. In general, we
```

scheme/Makefile
yogeshbdeshpande, line 1: This file, is not pertaining to your changes, perhaps, your changes needs re-base so that this file does not appear in this PR.

yogeshbdeshpande (commented):
started reviewing the changes, will complete it by tomorrow

scheme/Makefile
DhanusML: I assume the comment is on `/scheme/tpm-enacttrust/test/cmd/gen-token/main.go`. The change was done because copyright check (part of `make test`) was failing because the file was recently updated (in #449) but the year in the header was not updated.

store-plugin/corim-store/corimstore_enstore.go
DhanusML: package name should be a valid identifier, so it cannot contain `-` character (https://go.dev/ref/spec#Package_clause)

store-plugin/corim-store/corimstore_enstore.go
DhanusML: the struct `Store` should implement `IPluggable`, which has a method `GetAttestationScheme`. Right now, scheme name is only used to stub this method.

store-plugin/corim-store/corimstore_enstore.go
setrofim: The convention is to replace `-` with `_` when mapping directories to package names; so this should be `corim_store`

vts/cmd/vts-service/main.go
setrofim: The `go-plugin` should be renamed to `go-plugin-schemes` to be consistent with `go-plugin-stores`.

vts/cmd/vts-service/main.go
```diff
@@ -61,16 +68,33 @@ func main() {
...
+
+	endorsementStoreConfig, err := endorsementstore.AggregateStoreParams(
+		endorsementStorePluginConfig, subs["vts"], coservContext)
+
+	if err != nil {
+		log.Fatalf("could not create endorsement store configuration: %v", err)
+	}
+
+	// for builtin loader
+	builtinPluginConfig := maps.Clone(pluginConfig)
```
setrofim, line 93: I would drop `builtinPluginConfig` and just merge `endorsementStorePluginConfig` into `pluginConfig`, and later on, pass `pluginConfig` to all loaders.

We don't want builtin configuration to be in any way different from plugin configuration (the loading method should be as transparent as possible).

As in builtin configuration the same loader is used for both plugin types, it also means that the two types of plugins share a namespace (we cannot have a scheme plugin and a store plugin with the same name).  The loader will raise an error (though, as builtin loader is rarely exercised we might miss it), however when copying parameters below, rather an just using `maps.Copy()`, you need to make sure there are no clashing keys and raise an error.

An alternative, and I think this is the better solution, would be to ensure that store and scheme plugins live in different namespaces. This means using separate loaders, even in built in configuration. This would require modifying the builtin loader to take a list plugins as an argument rather than assuming `github.com/veraison/builtin.plugins` list. That list can than be split in two -- one for schemes and one for stores.

store-plugin/corim-store/corimstore_enstore.go
```diff
@@ -0,0 +1,148 @@
...
+
+func (s *Store) GetName() string {
+	return PluginName
+}
+
+func (s *Store) GetAttestationScheme() string {
+	return SchemeName
+}
+
+func (s *Store) GetSupportedMediaTypes() map[string][]string {
```
setrofim, line 140: So as discussed elsewhere, I don't think media types make sense the context of the store and this should be removed. 

The existing plugin code needs to be rejigged to not assume all plugins have associated media types. The simplest, bare minimum, solution would be to return `nil` here and make sure that gets interpreted correctly by the  loader.

A better solution would be to refactor the plugin code so that instead of just

```go
type GoPluginLoader struct {
	Location string

	logger            *zap.SugaredLogger
	loadedByName      map[string]IPluginContext
	loadedByMediaType map[string]IPluginContext

	pluginMap map[string]plugin.Plugin

	pluginParams map[string]*Parameters

	registeredPluginTypes map[string]string
}
```
that we have now, we would have

```go
type GoPluginLoader struct {
	Location string

	logger            *zap.SugaredLogger
	loadedByName      map[string]IPluginContext

	// This gets specified as Plugins when creating a new go-plugin client.
	pluginMap map[string]plugin.Plugin

	pluginParams map[string]*Parameters

	registeredPluginTypes map[string]string
}

type MediaTypeGoPluginLoader struct {
	GoPluginLoader

	loadedByMediaType map[string]IPluginContext
}
```

And all media type related methods moved into `MediaTypeGoPluginLoader` which would then correspond a new `IMediaTypeLoader` interface which contains and extends the existing `ILoader`. Similar modifications would need to be done to `IPluggable` and `IManager` interfaces and their implementations. So the media type functionally is layered on top  of general plugin functionality. The schemes will then use the new media type based API, and the stores can use the old API that no longer requires a vestigial media type.

vts/endorsementstore/config.go
```diff
@@ -0,0 +1,135 @@
...
+	"github.com/veraison/corim/comid"
+	"github.com/veraison/services/config"
+	"github.com/veraison/services/plugin"
+	vtscoserv "github.com/veraison/services/vts/coserv"
+)
+
+// AggregateStoreParams aggregates different configurations to create
+// a consolidated parameters map, that can be passed to the store plugin
+// manager. The configurations come from store plugin configuration
+// section, vts configuration section and coserv configuration section.
```
setrofim, line 20: Rather than doing this, I think it would make more sense to restructure configuration so that all endorsement-store related configuration is in a single section. In particular, the old CoSERV section does not need to exist

vts/endorsementstore/config.go
```diff
@@ -0,0 +1,135 @@
...
+// The configuration parameters common for all stores.
+// These parameters are extracted from VTS configuration and
+// broadcast to all store plugins.
+type StoreCommonParams struct {
+	FallbackAuthority *comid.CryptoKey
+	MaxExpiry         time.Duration
+}
+
+// Create StoreCommonParams from a coservContext
+func CreateStoreCommonParams(coservContext *vtscoserv.Context) (*plugin.Parameters, error) {
```
setrofim, line 85: This seems.. backwards. The context should be created from the config parameters and not the other way round.

vts/endorsementstore/store.go
```diff
@@ -0,0 +1,172 @@
...
+	if len(s.stores) == 0 {
+		return nil, ErrNoStores
+	}
+	return s.stores, nil
+}
+
+func CreateEndorsementStore(plugins []string, manager StoreManager, logger *zap.SugaredLogger) (handler.IEndorsementStore, error) {
+	if len(plugins) == 0 {
+		err := errors.New("no plugins in `active-stores` list")
+		logger.Errorf("could not create endorsementstore: %v", err)
```
setrofim, line 42: Do not log here. The error is returned and should be handled (including logging it) by the calling context.

vts/endorsementstore/store.go
```diff
@@ -0,0 +1,172 @@
...
+func (s *VtsEndorsementStore) storeList() ([]handler.IEndorsementStorePlugin, error) {
+	if len(s.stores) == 0 {
+		return nil, ErrNoStores
+	}
+	return s.stores, nil
+}
+
+func CreateEndorsementStore(plugins []string, manager StoreManager, logger *zap.SugaredLogger) (handler.IEndorsementStore, error) {
+	if len(plugins) == 0 {
+		err := errors.New("no plugins in `active-stores` list")
```
setrofim, line 41:
```suggestion
		err := errors.New("no names in `active-stores` list")
```

vts/endorsementstore/store.go
```diff
@@ -0,0 +1,172 @@
...
+}
+
+func (s *VtsEndorsementStore) storeList() ([]handler.IEndorsementStorePlugin, error) {
+	if len(s.stores) == 0 {
+		return nil, ErrNoStores
+	}
+	return s.stores, nil
+}
+
+func CreateEndorsementStore(plugins []string, manager StoreManager, logger *zap.SugaredLogger) (handler.IEndorsementStore, error) {
```
setrofim, line 39: rename `plugins` to `pluginNames`

vts/endorsementstore/config.go
DhanusML: coserv configuration should not be merged with the endorsement store
configuration because coserv service is the consumer of coserv configuration.
This pattern is followed in all the services except coserv section, which
contains max-expiry and signer sub-sections, which are used by vts. Maybe this
should be first separated as

```yaml
# used by coserv service
coserv:
  listen-addr: ...
  protocol: ...
  cert: ...
  cert-key: ...

# used by vts
coserv-signer:
  alg: ...
  key: ...
  max-expiry: ...
```
(expiry in the coserv-signer section because max-expiry is also (indirectly)
the validity period of the signature).

Now, the common coserv configuration which all the stores use are max-expiry and
the public part of signing key. Since vts is the primary owner of these
parameters, I decided to let vts parse these configurations and then extract
them from the signer object (coserv.Context) and pass them to the store.

It is possible to move just the expiry parameter to a unified store config, but
the fallback authority will have to be extracted from the signer.

vts/cmd/vts-service/config-docker.yaml
```diff
@@ -5,10 +5,13 @@ plugin:
   backend: go-plugin
   go-plugin:
     dir: ../../../scheme/bin/
```
thomas-fossati: this should go, right?

store-plugin/corim-store/corimstore_enstore.go
```diff
@@ -0,0 +1,146 @@
...
+	}
+	return res, err
+}
+
+func (s *Store) ExecuteCoservQuery(profile, query string) (*coserv.Coserv, error) {
+	// If reading CoSERV config failed during initialization,
+	// CoSERV interface would be disabled.
+	if s.CoservCfg == nil {
+		s.logger.Errorf("store is not configured for CoSERV")
+		return nil, errors.New("missing configurations for CoSERV service")
```
thomas-fossati, line 57: shouldn't this return one of the errors handled by `logOrErr()` (i.e., errnotfound/errunsupported)?  I think we want ExecuteCoservQuery to skip to the next active store rather than stop the loop.

store-plugin/corim-store/corimstore_enstore.go
```diff
@@ -0,0 +1,146 @@
...
+
+func (s *Store) AddCorimBytes(data []byte, label string, activate bool) error {
+	s.logger.Debugf("adding CoRIM")
+	return s.Store.AddBytes(data, label, activate)
+}
+
+func (s *Store) Fini() error {
+	s.logger.Info("closing corimstore")
+	if s.Store == nil {
+		panic("attempted to close an uninitialized store")
```
thomas-fossati, line 89: Is it necessary to panic here?  Can't we just return an error?

vts/trustedservices/trustedservices_grpc.go
```diff
@@ -619,29 +649,29 @@ func getEndorsementsError(err error) *proto.EndorsementQueryOut {
...
+		o.logger.Warnf("Bad request: could not parse media type: %v", err)
 		return nil, err
 	}
+	profile := mtParams["profile"]
 
-	return query.ToCBOR()
-}
+	resp, err := o.endorsementStore.ExecuteCoservQuery(profile, queryIn.Query)
+	if err != nil {
+		o.logger.Infof("could not find coserv result in store")
```
thomas-fossati: please attach the original error to the log message or it'll get lost.  (there may be other errors than ENOTFOUND that are worth surfacing.)

vts/endorsementstore/store.go
```diff
@@ -0,0 +1,171 @@
...
+}
+
+func (s *VtsEndorsementStore) AddCorimBytes(data []byte, scheme string, activate bool) error {
+	// only submit endorsements to the first store in list
+	stores, err := s.storeList()
+	if err != nil {
+		s.logger.Errorf("failed to fetch store list: %v", err)
+		return err
+	}
+	store := stores[0] // stores is guaranteed to contain at least one entry
```
thomas-fossati, line 134: question: how can we be sure that stores[0] is not read-only?

store-plugin/amd-kds-coserv/coserv_handler.go
```diff
@@ -50,7 +51,17 @@ func getVcekForInstance(instance *coserv.StatefulInstance) ([]byte, error) {
 	return certBytes, nil
 }
 
-func (s CoservProxyHandler) Init(*plugin.Parameters) error {
+func (s *CoservProxyHandler) Init(params *plugin.Parameters) error {
+	var cfg vtsstore.StoreCommonParams
+	if err := (&cfg).FromParams(params); err != nil {
+		s.StoreCommonParams = nil
```
thomas-fossati, line 59: please add a warning-level log here

thomas-fossati (commented):
a few more comments

store-plugin/corim-store/corimstore_enstore.go
setrofim: `Fini()` must only be called if `Init()` succeeded. Calling `Fini()` when the store is uninitialized is programmer error on behalf of the direct caller. `panic()` is appropriate here -- we don't want to propagate the error we want the call to `Fini()` to be removed/moved.

vts/cmd/vts-service/config-docker.yaml
DhanusML: Correct. `go-plugin` should be changed to `go-plugin-schemes`. Thanks for pointing it out. Same shoudl be done in `vts/cmd/vts-service/config.yaml`.

store-plugin/corim-store/corimstore_enstore.go
DhanusML: This is neither of those errors. Instead this is a misconfiguration because this happens when trying to execute coserv query when the store is not configured for coserv. This can be caught in `Init` by throwing an error if coserv config is not being passed but that would cause store initialization to fail on deployments that do not enable coserv service.

Note that `CoservCfg` is passed from vts to all store plugins if the config.yaml contains coserv configurations.

store-plugin/corim-store/corimstore_enstore.go
DhanusML: Since `Fini` is only called if `Init` was successful, `s.Store` cannot be `nil`. See  https://github.com/veraison/services/pull/435/changes#r3892508989

vts/endorsementstore/store.go
DhanusML: If `stores[0]` does not support write operation, in the current implementation, endorsement provisioning will fail. So the user should make sure that the first store in the list is not read-only. Since the default configurations had `corim-store` as the first entry, this was fine. Maybe a better approach would be to try `AddCorimBytes` until a store does not return `ErrUnsupported`.

vts/cmd/vts-service/main.go
```diff
@@ -31,8 +31,9 @@ func main() {
 		log.Fatalf("could not read config: %v", err)
 	}
 
-	subs, err := config.GetSubs(v, "store", "po-store",
-		"*po-agent", "plugin", "*vts", "ear-signer", "*coserv", "*logging", "*scheme")
+	subs, err := config.GetSubs(v, "po-store",
+		"*po-agent", "plugin", "*vts", "ear-signer", "*coserv-signer",
```
setrofim, line 35: I don't think the `coserv-signer` sub is ever used?

vts/endorsementstore/store.go
thomas-fossati: > If `stores[0]` does not support write operation, in the current implementation, endorsement provisioning will fail. So the user should make sure that the first store in the list is not read-only. Since the default configurations had `corim-store` as the first entry, this was fine.

ok.  Is this assumption clearly documented?

> Maybe a better approach would be to try `AddCorimBytes` until a store does not return `ErrUnsupported`.

definitely :-)

store-plugin/corim-store/corimstore_enstore.go
thomas-fossati: > This is neither of those errors. Instead this is a misconfiguration because this happens when trying to execute coserv query when the store is not configured for coserv.

ok, but if this is a misconfiguration, it shouldn't reach this deep.

> This can be caught in `Init` by throwing an error if coserv config is not being passed but that would cause store initialization to fail on deployments that do not enable coserv service.

tricky.

> Note that `CoservCfg` is passed from vts to all store plugins if the config.yaml contains coserv configurations.

so, is the error even possible?  If not, panic'ing is probably a better option.

vts/cmd/vts-service/main.go
```diff
@@ -42,10 +43,15 @@ func main() {
 		log.Fatalf("could not configure logging: %v", err)
 	}
 
-	log.Info("initializing stores")
-	enStore, err := store.New(subs["store"], log.Named("store"))
-	if err != nil {
-		log.Fatalf("endorsement store initialization failed: %v", err)
+	var coservContext *coserv.Context
+	if subs["coserv"].IsSet("signer") {
```
setrofim, line 47: <s>Actually, this entire part should probably be done inside the plugin.</s>

Nevermind; the remote store providers also need it.

vts/endorsementstore/store.go
setrofim: > > Maybe a better approach would be to try AddCorimBytes until a store does not return ErrUnsupported.

> definitely :-)

Actually, I'm not sure that's necessarily a great idea, since it's not immediately obvious where writes will be directed to. I think a better solution would be to ensure that the first (highest priority) store is writable. And rather than relying to catching and handling `ErrNotSupported`, the store API should be extended with a method for querying its capabilities (e.g. if it's writable); that way, this can be validated during initialization.

thomas-fossati: > > > Maybe a better approach would be to try AddCorimBytes until a store does not return ErrUnsupported.
> 
> > definitely :-)
> 
> Actually, I'm not sure that's necessarily a great idea, since it's not immediately obvious where writes will be directed to. I think a better solution would be to ensure that the first (highest priority) store is writable. And rather than relying to catching and handling `ErrNotSupported`, the store API should be extended with a method for querying its capabilities (e.g. if it's writable); that way, this can be validated during initialization.

that's an even better approach.  Though it's a bit more intrusive.

vts/endorsementstore/config.go
setrofim: > coserv configuration should not be merged with the endorsement store
configuration because coserv service is the consumer of coserv configuration.

Yes, you're right, sorry, I was over-eager with that because I was just looking at `vts/cmd/vts-service/config.yaml`.

> which contains max-expiry and signer sub-sections, which are used by vts. Maybe this should be first separated as [...]

Yes, moving that is probably the right thing to do. Actually, I think both `singer` and `max-expiry` can just move under `endorsement-store`; and `active-store` (currently under `vts`) should move there as well. This will remove the need for `AggregateStoreParams()` as everything will be in a single sub and can be loaded using the normal config loading mechanism.

EDIT: the names should probably change to make it clearer what the configs control; so the config would look something like

```yaml
endorsement-store:
  coserv-signer:
    alg: ES256
    key: ./skey.jwk
  coserv-max-expiry: 1m
  active-plugins:
    - corim-store
  plugin-parameters:
    corim-store:
      dbms: sqlite3
      dsn: file::memory:?cache=shared
      trace-sql: false
```

DhanusML: Rearranged the config for endorsement to something similar. Please take a look again.

vts/endorsementstore/store.go
DhanusML: Updating the plugin interface is the robust solution. I have created an issue to track that here: https://github.com/veraison/services/issues/472. Since anything else would be a workaround, I have kept this unchanged for now.  Also updated the readme in this directory to include the assumption.

store-plugin/corim-store/corimstore_enstore.go
DhanusML: The configuration parsing now ensures that if CoSERV configuration is not passed to the plugins, CoSERV interface must be disabled in VTS (vts/endorsementstore/config.go). Updated plugin to panic instead of returning error if this state is encountered.

shefali-kamal:
Hi @thomas-fossati @setrofim ,
we have addressed the review comments and gave created a GH Issue for other backlogs: https://github.com/veraison/services/issues/472
Please review the PR again.

vts/endorsementstore/Readme.md
```diff
@@ -0,0 +1,58 @@
...
+parameters are passed along with the plugin parameters to each of the plugins
+during plugin initialization.
+
+* `active-plugins`: contains the list of store plugins that will be used to
+construct the endorsement-store. The plugins will be iterated over in the
+order they are given in the config.
+
+* `plugin-parameters`: contains parameters for store plugins.
+
+Example:
```
setrofim, line 38: I think this is an old example that can now be removed?

setrofim (approved):

thomas-fossati (approved):
🚢 it!

shefali-kamal:
Hi @thomas-fossati,
we have resolved the merge conflicts. If this looks good please merge so that we can rebase ELM API pr and raise few more dependent PRs. Thanks again for all the help.

PR #441, opened 2026-08-25
fix: ignore nvidia scheme on systems without nvat.h

setrofim:
`nvidia` scheme is built against an external C library, which may or may
not be installed on host trying to build Veraison. To avoid requiring
everyone to worry about this, check whether the C library is already
installed, and stub-out make targets if it isn't. make will no longer
fail if the dependency is missing and will just generate a message
instead.

Since `nvidia` scheme is conditioned on this external dependency, remove
it from the built-in scheme list.

### DIFF
--- a/scheme/nvidia/check-nvat.sh
+++ b/scheme/nvidia/check-nvat.sh
@@ -1,0 +1,15 @@
+#!/usr/bin/env bash
+# Copyright 2026 Contributors to the Veraison project.
+# SPDX-License-Identifier: Apache-2.0
+
+if [[ "$(type -p cc)" == "" ]]; then
+	echo "C compiler not installed"
+	exit 1
+fi
+
+if ! (echo -e '#include <nvat.h>\n' | cc -E -x c - >/dev/null 2>&1); then
+	echo "nvat.h not found"
+	exit 1
+fi
+
+
--- a/builtin/schemes.go
+++ b/builtin/schemes.go
@@ -8,7 +8,6 @@ import (
 
 	scheme9 "github.com/veraison/services/scheme/amd-kds-coserv"
 	scheme3 "github.com/veraison/services/scheme/arm-cca"
-	scheme10 "github.com/veraison/services/scheme/nvidia"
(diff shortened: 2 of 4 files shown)


yogeshbdeshpande (approved):
LGTM!

yogeshbdeshpande:
@setrofim Please check why the CI is still failing?

setrofim:
> @setrofim Please check why the CI is still failing?

Missing copyright in `check-nvat.sh`. Fixed now.

scheme/nvidia/check-nvat.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+# Copyright 2026 Contributors to the Veraison project.
+# SPDX-License-Identifier: Apache-2.0
+
+if [[ "$(type -p cc)" == "" ]]; then
+	echo "C compiler not installed"
+	exit 1
+fi
+
+if ! (echo -e '#include <nvat.h>\n' | cc -E -x c - >/dev/null 2>&1); then
```
THS-on, line 10: Any reason not to use pkg-config?

setrofim: So it works on distros other than Ubuntu. Since this is for native build/test we cannot assume the distro  (e.g. I'm on Arch).

THS-on: I would assume that if it is packaged for other distros, they also ship the correct `.pc` file, pkg-config should be distro agnostic.

setrofim: Ah sorry,  you're right, it's not distro-specific. Any advantage to using pkg-config over the above? (the above seems more robust since it directly checks whether the header is found rather than relying on .pc)

yogeshbdeshpande:
@THS-on, shall we merge this then?

scheme/nvidia/check-nvat.sh
THS-on: True, not sure how nicely this plays with packaging, but let's merge this version.

THS-on (approved):

PR #445, opened 2026-08-26
build: add RPM release workflow

cowbon:
## Summary

- build an RPM whenever a GitHub Release is published
- upload the RPM as a release-versioned workflow artifact

## Testing

- YAML parsed successfully
- `git diff --check`

### DIFF
--- a/.github/workflows/release-rpm.yml
+++ b/.github/workflows/release-rpm.yml
@@ -1,8 +1,9 @@
-name: Build RPM release package
+name: Automatic RPM package creation based on the monthly tag
 
 on:
-  release:
-    types: [published]
+  push:
+    tags:
+      - v0.0.**
 
 permissions:
   contents: read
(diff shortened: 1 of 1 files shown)


setrofim:
Our current rpm build is Oracle-specific, we probably shouldn't create an rpm release until we have a distro-agnostic configuration.

cowbon:
Hi @setrofim, it's true that the current RPM spec file is designed for Oracle Linux. However, we do have debian specific configuration under deployment. Is there any blocker for enabling deb package building in Github workflows?

setrofim:
There is no blocker for deb package building. The deb deployment isn't Debian-specific -- it should work for any distro that uses deb packages (e.g. Ubuntu).

cowbon:
Actually Oracle Linux is RHEL-like, so in theory the same RPM can also be used for any Linux distro using RPM as the package manager

.github/workflows/release-rpm.yml
```diff
@@ -0,0 +1,59 @@
+name: Build RPM release package
+
+on:
+  release:
+    types: [published]
```
setrofim: We do not publish releases for `services`, instead, there is an automatic monthly tag. Packaging should be triggered off of that -- see `time-package.yml` workflow.

cowbon: Got it. Now the PR only brings in the workflow for building the RPMs like Debian packages, but not on every releases

PR #449, opened 2026-09-02
tpm-enacttrust: accept unpadded nonce encodings

flaviens:
Hi!

This PR is to accept more encodings for the Nonce, as requested in [https://github.com/veraison/services/pull/432.](https://github.com/veraison/services/pull/432.)
Please note that I used agents to proceed (although iteration through encodings makes sense to me).


yogeshbdeshpande (approved):
LGTM!

PR #451, opened 2026-09-02
Device Assignment (partial) attestation scheme

setrofim:
Implement an attestation scheme based on EAT-DA token containing SPDM measurements (legacy PCIe currently unsupported). This is intended to be used as part of composite attestation with CCA (binding TBD) and does not perform integrity validation on its own.

This also makes `GetTrustAnchorID()` optional for schemes.

### DIFF
--- a/handler/ischemehandler.go
+++ b/handler/ischemehandler.go
@@ -30,6 +30,12 @@ type ISchemeHandler interface {
 	// to the specified profile.
 	ValidateCorim(uc *corim.UnsignedCorim) (*ValidateCorimResponse, error)
 
+	// GetTrustAnchorIDs returns a slice of Environments used to
+	// retrieve the trust anchors associated with evidence. The trust
+	// anchors may be necessary to validate the entire evidence and/or extract
+	// its claims (if it is encrypted).
+	GetTrustAnchorIDs(evidence *appraisal.Evidence) ([]*comid.Environment, error)
+
 	// GetReferenceValueIDs returns a slice of Environments used to retrieve
 	// reference values for an attestation scheme, using the claims
 	// extracted from attestation token and the associated trust anchors.
--- a/handler/ischemeimplementation.go
+++ b/handler/ischemeimplementation.go
@@ -11,12 +11,6 @@ import (
 // ISchemeImplementation is the subset of the ISchemeHandler interface that
 // must be implemented by schemes that make use of SchemeHandlerWrapper.
 type ISchemeImplementation interface {
-	// GetTrustAnchorIDs returns a slice of Environments used to
-	// retrieve the trust anchors associated with evidence. The trust
-	// anchors may be necessary to validate the entire evidence and/or extract
-	// its claims (if it is encrypted).
-	GetTrustAnchorIDs(evidence *appraisal.Evidence) ([]*comid.Environment, error)
-
 	// ExtractClaims parses the attestation evidence and returns claims
 	// extracted therefrom. (Note: the integrity of the input evidence must
 	// be validated before any claims can be used; this can be done as part
--- a/handler/schemeimplementationwrapper.go
+++ b/handler/schemeimplementationwrapper.go
@@ -130,7 +130,15 @@ func (o *SchemeImplementationWrapper) ValidateCorim(uc *corim.UnsignedCorim) (*V
 func (o *SchemeImplementationWrapper) GetTrustAnchorIDs(
 	evidence *appraisal.Evidence,
 ) ([]*comid.Environment, error) {
-	return o.Impl.GetTrustAnchorIDs(evidence)
+	impl, ok := o.Impl.(interface {
+		GetTrustAnchorIDs(*appraisal.Evidence) ([]*comid.Environment, error)
+	})
+
+	if ok {
+		return impl.GetTrustAnchorIDs(evidence)
+	}
+
+	return nil, nil
 }
 
 func (o *SchemeImplementationWrapper) GetReferenceValueIDs(


scheme/da-spdm/REAME.md
```diff
@@ -0,0 +1,14 @@
+# Device Assignment Attestation for SPDX devices
+
+This scheme implements attestation based on EAT-DA tokens described by
+[draft-poirier-rats-eat-da]. Currently only SPDX devices are supported (legacy
```
thomas-fossati:
```suggestion
[draft-poirier-rats-eat-da]. Currently only SPDM devices are supported (legacy
```

scheme/da-spdm/test/corim/src/comid-da-bad-group-set.json
```diff
@@ -0,0 +1,39 @@
...
+    "version": 0
+  },
+  "triples": {
+    "reference-values": [
+      {
+        "environment": {
+          "class": {
+            "id": {
+              "type": "bytes",
+              "value": "c3BkbTpBQ01FOldJREdFVC1BOjAxMjM0NTY3ODk="
```
thomas-fossati, line 13: How is this obtained from evidence?

setrofim: This is the UTF-8 string containing `device-name` and treated as `[]bytes` (so the value you see is the base64 encoding of the actual name).

thomas-fossati: ok, in SPDM that's an instance id, though, not a class id.  It's the subject (SAN) of the leaf certificate.

setrofim: ok, I think I misinterpreted what the SAN was. Can we derive a class ID from it? e.g. by stripping the `CN` element (or the last part of the `ub-DMTF-device-info`? Or do you have an alternative suggestion for a class ID?

I don't thing we want the reference value lookup to be instance-based?

thomas-fossati: > ok, I think I misinterpreted what the SAN was. Can we derive a class ID from it? e.g. by stripping the `CN` element (or the last part of the `ub-DMTF-device-info`? Or do you have an alternative suggestion for a class ID?

Yes, I think it is possible.  The definition of `ub-DMTF-device-info` is:
```
UTF8String({
  DMTF-manufacturer":"DMTF-product":"DMTFserialNumber
})
```
where each component is guaranteed to not include `:`

So, `DMTF-manufacturer":"DMTF-product"` can be extracted unambiguously and provide a class-scoped identifier.

> I don't thing we want the reference value lookup to be instance-based?

yeah

setrofim: I made the update -- the class id now doesn't contain the CN element for SAN and the last :-spearated element for DMTF device info.

scheme/da-spdm/plugin/main.go
```diff
@@ -0,0 +1,16 @@
+// Copyright 2026 Contributors to the Veraison project.
+// SPDX-License-Identifier: Apache-2.0
+// Copyright <TODO> Contributors to the Veraison project.
```
yogeshbdeshpande: Why Line 3 & 4 and why it is <TODO>?

scheme/da-spdm/test/corim/src/comid-da-ref.json
```diff
@@ -0,0 +1,71 @@
+{
+  "tag-identity": {
+    "id": "00000000-0000-0000-DA54-000000000001",
+    "version": 0
+  },
+  "triples": {
+    "reference-values": [
```
yogeshbdeshpande, line 7: I am little puzzled by this example, as per the specification, 3.1.  SPDM Claims:

`A SPDM claim instance is expected to be present for each SPDM
   compatible device to be attested.  Each instance consists of a
   measurements section, a certificates section, or both.`

So when you changed the class-id in the measurement values map, does it mean this is a reference value for two independent devices???

yogeshbdeshpande: Seems a bit odd to have two independent device CoRIMs in one CoRIM Tag

yogeshbdeshpande (commented):
in middle of review, have some minor questions

scheme/da-spdm/test/corim/src/comid-da-ref.json
setrofim: I think so? I'm not sure what you mean by "you changed class-id in the measurement values map"; the class-id is in the environment. 

There are two value triples here, each containing a single measurement. Each triple has a different class-id identifying the SPDM device it pertains to. This matches the structure of the example EAT-DA token that has data on two SPDM devices as two submods. Th class-id is derived form the corresponding submod name.

So yes, there are two "devices", but they are components of a single attesting environment; hence they are grouped together. (Note: a far as the attestation is concern, "endorsements" are a sequence of value triples, so it doesn't really matter whether they're packaged as a single comid or multiple (or multiple corims even), however here, the structure of the endorsement matches the structure of the evidence on the assumption that the endorsements will be provided by the owner/manufacturer of the device comprising the attesting environment, rather than by manufacturer of the individual SPDM devices).

scheme/da-spdm/REAME.md
atulfj: filename typo: README.md?

scheme/da-spdm/scheme.go
```diff
@@ -0,0 +1,436 @@
...
+	}
+
+	return common.ToMapViaJSON(claims)
+}
+
+func (o *Implementation) AppraiseClaims(
+	claims map[string]any,
+	endorsements []*comid.ValueTriple,
+) (*ear.AttestationResult, error) {
+	claimsTripleMap, err := clamsToTripleMap(claims)
```
atulfj: typo: claimsToTripleMap

scheme/da-spdm/scheme.go
```diff
@@ -0,0 +1,436 @@
...
+
+		for _, part := range parts {
+			part = strings.TrimSpace(part)
+			if strings.HasPrefix(part, "CN=") {
+				continue
+			}
+			out = append(out, part)
+		}
+
+		out[0] = fmt.Sprintf("spdm:%s", out[0])
```
atulfj: If all the SANs are of the form "spdm:CN=01234", does this cause a panic, since `out` will be empty?

scheme/da-spdm/scheme.go
```diff
@@ -0,0 +1,436 @@
...
+	return triplesMap, nil
+}
+
+func endorsementsToTripleMap(endorsements []*comid.ValueTriple) (map[string]*comid.ValueTriple, error) {
+	ret := make(map[string]*comid.ValueTriple)
+
+	for i, endorsement := range endorsements {
+		submodName, err := submodNameFromEnvironment(&endorsement.Environment)
+		if err != nil {
+			return nil, fmt.Errorf("endrosement[%d]: %w", i, err)
```
atulfj: typo: endorsement

scheme/da-spdm/plugin/main.go
setrofim: This was left over from the example "scheme"; now fixed.

scheme/da-spdm/test/corim/src/comid-da-ref.json
yogeshbdeshpande: Apologies, I messed up the issues, so deleting the accidental collision

yogeshbdeshpande: I meant changing the class-id in the class-map. I will look at the EAT DA spec and update the thread accordingly.!!

yogeshbdeshpande:
@setrofim I will complete my review before Monday - the 07th !

scheme/da-spdm/test/corim/src/comid-da-ref.json
yogeshbdeshpande: After checking the EAT DA specification, if you are referring to the Overall signature, then it is coming from the Lead Attester (LA), so this does not correspond to the Individual Device Signatures. LA is responsible for integrity of the overall token, but here we are referring to Reference Values of two independent devices, coming from two separate sources.

In theory they may come from the same Endorser but in practice, this is not a correct illustration.

Also note the Individual SPDM Claims may be signed and the signature is from Attesting Env. of each of them which will be seperate.

scheme/da-spdm/scheme.go
```diff
@@ -0,0 +1,436 @@
...
+		}
+
+		if matchResult.Matched {
+			appraisal.TrustVector.Executables = ear.ApprovedRuntimeClaim
+		 } else {
+			appraisal.TrustVector.Executables = ear.UnrecognizedRuntimeClaim
+		}
+	}
+
+	// remove the scheme submod from the result, as it it is not used --
```
yogeshbdeshpande:
```suggestion
	// remove the overall attestation scheme name specific submod from the result, as it is not used --
```

scheme/da-spdm/scheme.go
```diff
@@ -0,0 +1,436 @@
...
+			continue
+		}
+
+		submodClaimsMap, err := common.ToMapViaJSON(submodClaims)
+		if err != nil {
+			return nil, fmt.Errorf("submod[%s]: %w", submodName, err)
+		}
+		appraisal.VeraisonAnnotatedEvidence = &submodClaimsMap
+
+		matchResult, err := matchMeasurements(
```
yogeshbdeshpande, line 151: At the moment, this scheme only matches Measurement claims

is there a plan to support `Certificate Claims` in the next iteration?

yogeshbdeshpande (approved):
In general LGTM, I have completed my review with some comments.

thomas-fossati (approved):
LGTM, thanks!

Issue #192, opened 2023-09-05
HTTP/REST API code restructure


setrofim:
The code for each REST node (provisioning, verification, etc) is contained entirely within it's own sub-package. This leads to significant duplication of common or similar functionality (e.g. `ReportProblem` is re-implmented inside each sub-package).

We should fuctor out `ReportProblem` along with some other common bits (which may require some additionaly parametrisation) into a separate sub-package which is then re-used by all node implementations.

setrofim:
Closing as realistically this will only happen as part of some wider refactor tied into something else.

setrofim:
Reopenning due to this discussion: https://github.com/veraison/services/pull/453#discussion_r4036116592

PR #453, opened 2026-09-04
feat: implement endorsements lifecycle mgmt. API

shefali-kamal:
This PR implements the REST API endpoints and handlers for activation and deactivation of endorsement triples as specified by the spec (See
https://github.com/veraison/docs/blob/main/api/endorsement-provisioning/endorsement-provisioning.yaml).

- update some copyrights
- use application/concise-problem-details+cbor
- add integration-tests

Addresses Issue: https://github.com/veraison/services/issues/402


integration-tests/scripts/gen-elm-queries.sh
setrofim, line 1: "ELM" is very non-obvious. This acronym is not explained anywhere, and even if it was, it's not widely-used enough to be recognized. Favour clarity over brevity when naming things. So something like `gen-life-cycle-management-queries.sh`

integration-tests/scripts/gen-elm-queries.sh
```diff
@@ -0,0 +1,64 @@
...
+# Copyright 2026 Contributors to the Veraison project.
+# SPDX-License-Identifier: Apache-2.0
+
+set -o pipefail
+set -eu
+
+A=${A?must be set to one of rv or ta or rim}
+O=${O?output file}
+
+# ref-value ELM query
```
setrofim, line 11: Similar to the above comment, "ELM" should be expanded here 
```suggestion
# ref-value Endorsement Life-cycle Management (ELM) query
```

ditto in comments below.

integration-tests/tests/test_cca_end_to_end.tavern.yaml
```diff
@@ -24,9 +24,13 @@ marks:
         - nonce
         # Indicates whether to use signed or unsigned CoRIM
         - corim_type
+        # Indicates whether to use rim selector or environment selector in ELM query
```
setrofim, line 27:
```suggestion
        # Indicates whether to use rim selector or environment selector in Endorsement
        # Life-cycle Management query
```

integration-tests/utils/generators.py
```diff
@@ -318,8 +316,23 @@ def substitute_random_corim_id(path):
...
-    data['corim-id'] = str(uuid.uuid4())
+    new_corim_id = str(uuid.uuid4())
+    data['corim-id'] = new_corim_id
 
     with tempfile.NamedTemporaryFile(delete=False, mode='w') as tf:
         json.dump(data, tf)
-        return tf.name
+        return tf.name, new_corim_id
+
+def generate_elm_queries(test):
```
setrofim, line 326: As elsewhere -- expand "elm" in the name -- clarity over brevity.

vts/trustedservices/trustedservices_grpc.go
```diff
@@ -838,3 +843,301 @@ func SerializeCertPEMBytes(certPEMs [][]byte) ([]byte, error) {
...
+			return activateEndorsementsErrorResponse(err), nil
+		}
+	case elmq.RimSelector != nil:
+		if err := o.activateEndorsementsUsingIDs(&elmq, req.SetActive); err != nil {
+			return activateEndorsementsErrorResponse(err), nil
+		}
+	default:
+		return activateEndorsementsErrorResponse(
+			errors.New("invalid ELM query"),
+		), nil
```
setrofim, line 874:
```suggestion
		// should not get here -- query already validated
		panic("invalid ELM query")
```

vts/trustedservices/trustedservices_grpc.go
```diff
@@ -838,3 +843,301 @@ func SerializeCertPEMBytes(certPEMs [][]byte) ([]byte, error) {
...
+		// by seeing if the ID parses as a valid UUID.
+		// see: https://github.com/veraison/corim-store/blob/d1c9b858ead512ff45ba54d0e16781fafcff242c/pkg/model/manifest.go#L94
+		var tagIdType model.TagIDType
+		if _, err := uuid.Parse(tagId); err == nil {
+			tagIdType = model.UUIDTagID
+		} else {
+			tagIdType = model.StringTagID
+		}
+
+		log.Infof("tag id: %v, tag id type: %v", tagId, tagIdType)
```
setrofim, line 1051:
```suggestion
		log.Debugf("tag id: %v, tag id type: %v", tagId, tagIdType)
```

vts/trustedservices/trustedservices_grpc.go
atulfj: But is it a good idea to `panic` at runtime?

setrofim: As per the comment, the `panic()` indicates that this case should be unreachable, so it won't panic at runtime. If, for some reason, it ever does, that is an indication of a programming problem in the immediate code above, and not an error that needs to be handled by the caller.

integration-tests/scripts/gen-elm-queries.sh
yogeshbdeshpande: Agree with Sergei on this one

integration-tests/scripts/gen-lifecycle-management-queries.sh
atulfj: Thanks for the review. I have expanded ELM and used more descriptive names where required.

vts/trustedservices/trustedservices_grpc.go
atulfj: Hmm, updated now as per your suggestion

provisioning/api/handler.go
```diff
@@ -140,6 +142,65 @@ func (o *Handler) Submit(c *gin.Context) {
 	sendSuccessfulProvisioningSession(c)
 }
 
+func (o *Handler) Activate(setActive bool) gin.HandlerFunc {
```
setrofim, line 145: Rename to `SetActive()`, and similarly for the GRPC and other API. "Activate" implies that you're, well, activating.  So when reading `handler.Activate(false)`, it's not obvious that that means "deactivate".

atulfj: updated

provisioning/lifecycle/query.go
```diff
@@ -0,0 +1,159 @@
...
+	"fmt"
+
+	"github.com/fxamacker/cbor/v2"
+	"github.com/veraison/corim/coserv"
+	"github.com/veraison/corim/encoding"
+	"github.com/veraison/eat"
+	"github.com/veraison/swid"
+)
+
+type Query struct {
```
setrofim, line 15: I think it's better if this doesn't clash unnecessarily with the CoSERV `Query`.  Maybe define it as

```golang
type Query struct {
    coserv.Query

    Profile *eat.Profile `cbor:"265,keyasint,omitempty" json:"profile,omitempty"`
}
```
and then have some additional validation to make sure `ResultType` is not set. 

At the very least, the fields that the two queries have in common should have the same CBOR and JSON keys.

shefali-kamal:
@setrofim we have addressed the review comments. Please check

provisioning/lifecycle/query.go
atulfj: I don't see why they will clash; they're two different types of `Query` in separate namespaces. 

It's not a good idea to embed `coserv.Query` inside `lifecycle.Query`. They have some fields in common, but that is where the similarity ends. Right now, it will be simple to validate that `ResultType` is not set, but it takes away the freedom from `lifecycle.Query` to evolve separately from `coserv.Query`. We do not want to keep adding more validation checks if `coserv.Query` changes in future.

CBOR keys are different, but I don't see how that poses any significant problem. And there are no JSON keys in `coserv.Query`.

setrofim: > They have some fields in common, but that is where the similarity ends.

I disagree. They're both queries used to select triples/corims from some underlying store. While they are used in different contexts, and so have some differing fields, the similarity is more than incidental -- conceptually, they do the same thing: select triples/corims. 

> but it takes away the freedom from lifecycle.Query to evolve separately from coserv.Query. 

That's kind of the point. I _want_ that freedom taken way, unless it's absolutely needed -- I'd rather not see two query formats diverge when they don't have to. This is why I suggested embedding -- it minimizes the risk of the two queries adding functionally similar fields under different keys in the future.

 If you'd rather avoid the validation complications that come with embedding, that's fine, but the key values should still be aligned; and I'd leave a comment here referencing `coserv.Query`. That way, when someone wants to extend this in the future, they'd at least have something prompting them to have a look at the other structure to make sure this remains aligned in case the other query was already extended in a similar way.

> CBOR keys are different, but I don't see how that poses any significant problem. And there are no JSON keys in coserv.Query.

The problem is the unnecessary confusion this will cause during debugging, when you see similar structures in CBOR dumps/diags with jumbled up fields and have to remember which key should correspond to which field in which context. This may not be a _huge_ deal, but on the other hand, the cost to staying consistent is virtually nil; so why diverge?

vts/trustedservices/trustedservices_grpc.go
```diff
@@ -837,3 +841,300 @@ func SerializeCertPEMBytes(certPEMs [][]byte) ([]byte, error) {
 
 	return allPEM.Bytes(), nil
 }
+
+func (o *GRPC) SetEndorsementsActive(
```
thomas-fossati, line 845: lacking a better place, I am leaving the comment here: I am confused by the absence of per-tenant scoping of the API.  Am I missing something?

vts/trustedservices/trustedservices_grpc.go
```diff
@@ -837,3 +841,300 @@ func SerializeCertPEMBytes(certPEMs [][]byte) ([]byte, error) {
...
+	keyTripleQueryGroup *corimstore.KeyTripleQueryGroup,
+	setActive bool,
+) error {
+	txStore, err := o.Store.BeginTx(nil)
+	if err != nil {
+		return err
+	}
+
+	if valueTripleQueryGroup != nil {
+		if _, err := txStore.SetValueTriplesActive(valueTripleQueryGroup, setActive); err != nil {
```
thomas-fossati, line 1109: checking: is it safe/correct to ignore an empty result set?

vts/trustedservices/trustedservices_grpc.go
```diff
@@ -837,3 +841,300 @@ func SerializeCertPEMBytes(certPEMs [][]byte) ([]byte, error) {
...
+			_ = txStore.Tx().Rollback()
+			return err
+		}
+	}
+
+	if keyTripleQueryGroup != nil {
+		if _, err := txStore.SetKeyTriplesActive(keyTripleQueryGroup, setActive); err != nil {
```
thomas-fossati, line 1116: ditto

provisioning/api/common.go
```diff
@@ -22,3 +25,26 @@ func ReportProblem(c *gin.Context, status int, details ...string) {
...
+	}
+
+	logger := log.Named("api")
+
+	log.LogConciseProblem(logger, status, prob)
+
+	b, err := cbor.Marshal(prob)
+	if err != nil {
+		log.Error(logger, "failed to marshal problem details to CBOR", "error", err)
+		c.AbortWithStatus(status)
```
thomas-fossati, line 71:
```suggestion
		c.AbortWithStatus(status)
		return
```

log/log.go
```diff
@@ -496,3 +496,24 @@ func LogProblem(logger *zap.SugaredLogger, prob *problems.DefaultProblem) {
 
 	logFunc("problem encountered", "title", prob.Title, "detail", prob.Detail)
 }
+
+// ConciseProblem is a representation of the problem details structure defined
```
thomas-fossati: This is already implemented in `coserv/api/handler.go::reportProblem()`

provisioning/api/handler.go
```diff
@@ -143,6 +145,65 @@ func (o *Handler) Submit(c *gin.Context) {
...
+			ReportConciseProblem(
+				c,
+				http.StatusUnsupportedMediaType,
+				fmt.Sprintf("unknown media type: %v, expected: %v", mediaType, ELMQueryMediaType),
+			)
+			return
+		}
+
+		// read body
+		payload, err := io.ReadAll(c.Request.Body)
```
thomas-fossati: Shouldn't we use `http.MaxBytesReader`?

provisioning/api/handler.go
```diff
@@ -125,7 +127,7 @@ func (o *Handler) Submit(c *gin.Context) {
 	if err != nil {
 		o.logger.Errorw("submit endorsement failed", "error", err)
 
-		if errors.Is(err, errors.New("no connection")) {
+		if strings.Contains(err.Error(), "no connection") {
```
thomas-fossati: this could test `vtsclient.NoConnectionError` directly

provisioning/lifecycle/query.go
atulfj: > conceptually, they do the same thing: select triples/corims.

This is true, and that is why `EnvironmentSelector` and `RimSelector` are entirely borrowed from coserv.

>If you'd rather avoid the validation complications that come with embedding, that's fine, but the key values should still be aligned; and I'd leave a comment here referencing coserv.Query. That way, when someone wants to extend this in the future, they'd at least have something prompting them to have a look at the other structure to make sure this remains aligned in case the other query was already extended in a similar way.

This assumes that `coserv.Query` will always be ahead of `lifecycle.Query`. It is possible that some features are needed in the latter that do not make much sense for the former. If `coserv.Query` is embedded, I need to go there and make the change, which does not sound good. Especially when CoSERV query is an IETF standard and ELM query is somewhat less serious.

For example, having a JSON serialization is more useful for `lifeycle.Query` than for `coserv.Query`, because it allows vendors/provisioners to create queries from JSON templates. [An issue I raised](https://github.com/veraison/corim/issues/282) was related to this concern, where I requested for `EnvironmentSelector` to support JSON tags. 

This is my primary concern, otherwise I'm on board with your idea.

log/log.go
atulfj: You're right, and the `reportProblem()` function is also same as `provisioning/api/common.go:ReportConciseProblem()`. But `reportProblem()` cannot be used as it's unexported. 
Should we put this stuff in a common place so that it can be referenced by all?

vts/trustedservices/trustedservices_grpc.go
atulfj: I'm a bit unfamiliar with what "per-tenant scoping" means. I noticed tenant IDs being used in some calls and log statements but unclear on how it actually affects the operations. Can you describe this or point me to a reference?

provisioning/lifecycle/query.go
setrofim: > This is true, and that is why EnvironmentSelector and RimSelector are entirely borrowed from coserv.

Exactly; so as well as re-using the field types, you should re-use the field keys (for the same reason that you're re-using the types).

> If coserv.Query is embedded, I need to go there and make the change, which does not sound good. 

Only if it's warranted for `coserv.Query` (e.g. if you want to add a new field which is not relevant to `coserv` you just add it to the embedding struct), and in that case, yes, that's exactly what you should do; this is the point -- we want to avoid accidental divergence. Note that even without embedding you'd still have to do that since you re-using `coserv` types for the fields anyway.

>  Especially when CoSERV query is an IETF standard and ELM query is somewhat less serious.

Which is why the burden is on the ELM query to stay current with `coserv` and not other way round. 

> For example, having a JSON serialization is more useful for lifeycle.Query than for coserv.Query, because it allows vendors/provisioners to create queries from JSON templates. https://github.com/veraison/corim/issues/282 was related to this concern, where I requested for EnvironmentSelector to support JSON tags.

You're kinda proving my point here. You're re-using the `coserv` types (which is absolutely the Right Thing to do, rather than re-implementing them); so the burden to keep in sync already exists to an extent. All embedding would do is extend it to the outer struct fields.

Anyway, I am fine with not embedding as long as

- the CBOR keys are aligned
- There a comment linking this to `coserv.Query`

Embedding would just mean we won't have to rely on future maintainer reading the comment, which would be preferable.

log/log.go
setrofim: But we don't want `provisioning` to be importing from `coserv`; Though thinking about it, this doesn't belong in `log` either since this is nothing to do with logging and is related to REST APIs instead.

This is actually a common problem with our frontends -- we don't want them to depend on each other so they end up duplicating a lot of functionality. We used to have an issue to address this (https://github.com/veraison/services/issues/192) that was closed  after being for ages because it seemed that we would never actually get around to it. Maybe it's worth re-openning?

Anyway, short-term, this should probably move into `provisioning` rather than be in 'log'.

EDIT (sorry, @atulfj, I didn't referesh before posting and didn't see your reply):

> Should we put this stuff in a common place so that it can be referenced by all?

That would be the right thing to do,  though that would be a bigger change affecting the other front-ends (and other API probably -- actually need evaluate how much can be make common). That is best done as a separate refactor, rather than as part of this pull.  For now, just move this out of `log` into `provisioning`.

atulfj: `provisioning` is a reasonable place, I'll move it there.

I vote for re-opening the issue and doing it incrementally instead of in one big sweep. We can start with moving just `ReportProblem()` to a common place and let other such stuff settle in over time.

vts/trustedservices/trustedservices_grpc.go
setrofim: It's not being ignored; the store will return `ErrNoMatch` if the query failed to match. The front-end will return `http.BadRequest`.

provisioning/api/handler.go
```diff
@@ -143,6 +145,65 @@ func (o *Handler) Submit(c *gin.Context) {
...
+			if setActive {
+				action = "activate"
+			}
+
+			o.logger.Errorw(action+" endorsement failed", "error", err)
+
+			if strings.Contains(err.Error(), "no connection") {
+				ReportConciseProblem(c,
+					http.StatusInternalServerError,
+					err.Error(),
```
setrofim: Do not insert the actual error into the response; that might leak internal server state. Log the actual error here , and in the response, set something generic like "problem  updating the store".

provisioning/api/handler.go
```diff
@@ -143,6 +145,65 @@ func (o *Handler) Submit(c *gin.Context) {
...
+		err = o.Provisioner.SetEndorsementsActive(tenantID, payload, setActive)
+		if err != nil {
+			action := "deactivate"
+			if setActive {
+				action = "activate"
+			}
+
+			o.logger.Errorw(action+" endorsement failed", "error", err)
+
+			if strings.Contains(err.Error(), "no connection") {
```
setrofim: I don't think this is a safe thing to do; it's also possible that there could be database-related errors that aren't related to the input (e.g. if the db hasn't been migrated).

I think it's better to invert the logic and detect specific errors (such as `ErrNoMatch`) to be reported as `BadRequest` and to report everything else  as `InteranalServerError`.

log/log.go
setrofim: > We can start with moving just ReportProblem() to a common place and let other such stuff settle in over time.

I specifically want to avoid doing this piecemeal and instead do this as a single logically-coherent refactor. That way things are less likely to fall through the cracks.

provisioning/api/handler.go
atulfj: `Submit()` also has this issue. Should I change in all places?

setrofim: Yes, thank you (please do changes unrelated to lifecycle management as a separate commit).

log/log.go
thomas-fossati: I don't see a problem with taking a piecemeal approach, but I also don't want to block this PR on a refactoring issue.
As long as it's tracked and we promise not to forget about it this time 😄, I'm happy.

setrofim: I've reopened the issue; so we're now tracking this.

vts/trustedservices/trustedservices_grpc.go
setrofim: Our multi-tenant support is somewhat vestigial at the moment, but we do try to maintain the plumbing for it. In the other API this is done vial labels that are constructed from the combination of tenant id (currently hard-coded to a dummy value) and the attestation scheme name derived fromm the media type (e.g. https://github.com/veraison/services/blob/main/vts/trustedservices/trustedservices_grpc.go#L274).

However, in this case, the media type just identifies the query, and not a specific scheme; so that approach won't work. Fully supporting this will require updating `corim-store` to either allow pattern matching for the label or support tenant ID as a distinct field (in retrospect, bundling it into label may not have been the best idea).

For now, however, you can at least make sure that the tenant ID gets propagated from the front-end to the VTS, the way it is done for the other API.

thomas-fossati: > For now, however, you can at least make sure that the tenant ID gets propagated from the front-end to the VTS, the way it is done for the other API.

👍 That sounds like a sensible first step.

Let's also add a tracking issue to veraison/corim-store for the

> updat[e] corim-store to either allow pattern matching for the label or support tenant ID as a distinct field (in retrospect, bundling it into label may not have been the best idea).

part

setrofim: done: https://github.com/veraison/corim-store/issues/29

provisioning/api/handler.go
atulfj: Catching `ErrNoMatch` is tricky because VTS only returns string-form error details in a `proto/status.proto` structure. The best we can do for matching is to compare the strings. Should I go ahead with it or create a better mechanism using grpc's status codes by importing the structure they provide: [grpc/status/status.proto](https://github.com/grpc/grpc/blob/420d61168e4f5a96d8679bfa92d69be9f59f5e45/src/proto/grpc/status/status.proto#L80)?

setrofim: String matching should be fine for now (that's what you're currently doing anyway). IIRC `ErrNoMatch` is always a fixed string.

provisioning/api/handler.go
atulfj: thanks, using it now

provisioning/api/handler.go
atulfj: yep, updated

log/log.go
atulfj: I've moved both `ConciseProblem` struct and `LogConciseProblem()` to `provisioning` (to avoid an import cycle with `log`). Added a comment there as a reminder to restructure.

provisioning/api/handler.go
atulfj: Done

provisioning/api/handler.go
atulfj: Done for now.

But it seems a broad issue - even in `Submit()`, we need to infer whether the error was because of the corim being malformed or because of an internal error. If we have a catch-all `InternalServerError`, the information is not enough (or correct) for the API user to know what the real problem was. So for now, I've avoided inverting the current logic.

provisioning/api/common.go
atulfj: added

provisioning/lifecycle/query.go
atulfj: `coserv.Query` is now embedded in `lifecycle.Query`. Updated JSON-related stuff and the above-mentioned issue accordingly.

setrofim (approved):

thomas-fossati (approved):
🚢 it, thanks!

yogeshbdeshpande:
@shefali-kamal I will review it by tomorrow.

PR #455, opened 2026-09-08
build: install libnvat in the vts and management images

bpayne-fstk:
## Problem

The NVIDIA GPU scheme plugin `dlopen`s `libnvat.so.1` at runtime, but neither
the `vts` nor the `management` image carries it. The plugin therefore fails to
load:

```
libnvat.so.1: cannot open shared object file: No such file or directory
```

vts logs this and continues *without* the plugin, so NVIDIA GPU evidence is
rejected as an unsupported media type. The failure is easy to miss because
nothing else about the deployment looks wrong — the service starts and serves
every other scheme normally.

## Change

`libnvat` is installed from NVIDIA's CUDA repository in the two images that
load the scheme plugin. The other service images are untouched.

Those two images move from `debian` to `ubuntu:24.04`, because NVIDIA publishes
`libnvat` for Ubuntu but not for Debian Trixie. I would rather not have changed
the base image, and I am happy to rework this if you would prefer a different
approach — installing into Debian from the Ubuntu repository seemed worse than
matching the distribution NVIDIA actually targets.

## Notes

* The repository path and architecture are derived rather than hardcoded: the
  distro tag comes from `/etc/os-release`, and `dpkg --print-architecture`
  selects `x86_64` or `sbsa`. The same Dockerfile builds on amd64 and arm64
  hosts, matching the arm64 runners added in fb400be.
* An unrecognised architecture fails the build rather than silently producing an
  image without the library.
* `wget` is purged in the layer that installs it, so it does not persist into
  the final image.

## Testing

Built both images for `linux/amd64` and confirmed the library is present and
resolvable:

```
$ docker run --rm --entrypoint sh veraison/vts -c 'ls -l /usr/lib/*/libnvat.so*'
lrwxrwxrwx 1 root root       16 /usr/lib/x86_64-linux-gnu/libnvat.so.1 -> libnvat.so.1.2.2
-rw-r--r-- 1 root root 23177448 /usr/lib/x86_64-linux-gnu/libnvat.so.1.2.2
```

The other four service images were rebuilt and confirmed unchanged on Debian
Trixie.


setrofim:
@bpayne-fstk please sign off the commit.

bpayne-fstk:
Signed off — apologies for missing that.

I also found why the integration tests were failing, and it was this change's fault rather than a flake. Moving these two images to `ubuntu:24.04` silently moved the `veraison` user's uid: `useradd --system` takes the next free system uid, which is **997 on debian trixie and 999 on ubuntu 24.04** (verified by inspecting both built images). The gid was already pinned at 616; the uid never was, so it had been tracking the base image.

Since the deployment shares the stores and logs volumes across services, that left vts and management owning files as 999 while provisioning, verification and coserv stayed at 997 — so anything one side created without group write was unusable to the other. The symptom matched: 14 of 17 tavern tests failed with

```
could not check media type with provisioner: rpc error: code = Unavailable
```

which is what the other services report when vts isn't serving, and the container-logs artifact contained no vts log at all.

Fixed by pinning `-u 997`, so every service image sits at the same uid whatever base it uses and this change no longer has an effect outside the two files it touches. Confirmed 997 is free on `ubuntu:24.04` and that pinning yields `uid=997(veraison) gid=616(veraison)`.

Worth saying plainly: I'd only tested these images on Kubernetes, where the config logs to stdout and there is no shared logs volume, so the uid shift was invisible to me. Your integration tests caught a real defect.

setrofim (approved):

PR #457, opened 2026-09-09
chore: remove riot scheme

setrofim:
This scheme was based on an early, incomplete draft of Microsoft RIOT, which has since itself been shelved.

At this point, this attestation scheme is not only not usable (it never was), but doesn't even serve as a good example of anything or a basis to develop on top of.

There is no reason to keep it in the code line.


yogeshbdeshpande:
@setrofim , shouldn't the code has the basics to implement either a `tcg-dice` scheme, or a profile of `tcg-dice` and/or open-dice..?

I agree on a standalone, it has no use.

setrofim:
Not really. RIOT did not use TCG DICE structures, so the two had little in common, aside form some general DICE concepts of layered attestation and that both are based on X.509 certs (though used them in very different ways).

THS-on (approved):

PR #459, opened 2026-09-10
fix(end-to-end): update CoRIM validity

setrofim:
Update validity inside input comments, as existing not-after was in
2025. Setting to 2099 so we don't have to worry about it for the next 73 years.

Since the attestation results printed by evcli are base64 encoded, the issue isn't immediately apparent, but anyone who unpacks the results will see it as "contraindicated", and may assume there is an issue with the deployment.


yogeshbdeshpande (approved):
LGTM!

yogeshbdeshpande:
@setrofim Do we need the same change at all places, example `evcli`, rust-corim, cover etc..?

setrofim:
Potentially. I have no idea where the JSON originally came from.

PR #468, opened 2026-09-14
fix(auth): opaque basic auth failures

setrofim:
Modify basic authorization flow so that failed authorization does not reveal information about the existence of the specified user name.

- The same problem detail is used regardless of whether it was the user name or the password that failed to match.
- Bcrypt hash comparison executes against a dummy value if a user name is not recognised; this prevents timing information from being used to infer the existence of a user name.

Add tests and auth to the overall make flow (so that tests run with "make test"). Add auth to coverage exceptions, as no other unit tests exist yet (thus far, auth has been tested via integration tests).


auth/basic.go
```diff
@@ -103,30 +103,38 @@ func (o *BasicAuthorizer) GetGinHandler(role string) gin.HandlerFunc {
...
 		if !hasAuth {
-			c.Writer.Header().Set("WWW-Authenticate", "Basic realm=veraison")
-			ReportProblem(c, http.StatusUnauthorized,
-				"no Basic Authorizaiton given")
-			return
+			o.logger.Warn("request does not contain basic auth")
+			haveProblem = true
 		}
 
 		userInfo, ok := o.users[userName]
```
THS-on: This should be probably a iteration over all users with a const compare.

auth/basic.go
```diff
@@ -103,30 +103,38 @@ func (o *BasicAuthorizer) GetGinHandler(role string) gin.HandlerFunc {
...
-				"no Basic Authorizaiton given")
-			return
+			if !haveProblem {
+				o.logger.Warnw("basic auth: unknown user", "user", userName)
+				haveProblem = true
+			}
+
+			userInfo = &basicAuthUser{
+				// dummyPassword
+				Password: "$2b$12$jul9hCKZP4cOF0hul0pmguzaJKLPfJ477NglE526eSYHUu9Rqe3UG",
```
THS-on, line 133: Do we assume the cost of the bcrypt is the same for all passwords?

THS-on (commented):
Two small things. Thanks for taking this

auth/basic.go
setrofim: Why?

THS-on: Otherwise you have a timing attack if a user exists or not.

auth/basic.go
setrofim: It's the same for all comparisons to a given hash. Whether all hashes have the same cost is not something we can control, since they're read from config and are provided by the user.

auth/basic.go
setrofim: The map lookup is constant time regardless of whether the key is in map or not, right?

THS-on: Does go guarantee this? I could not find anything. Just as comparison how Gin implements it: https://github.com/gin-gonic/gin/blob/master/auth.go

For small lists this probably does not matter.

setrofim: It may  vary under heavy collision, but that isn't a concern given  that this implementation is only designed for a small number of users  (they're read from the config file, rather than an actual user database).

THS-on: Makes sense. Then let's add this explicitly in the documentation and and as a comment (though we are already saying DO NOT use this in production), that these are not guaranteed to be timing side channel free. 

So that we hopefully do not get duplicated reports on this.

setrofim: I don't think it's that bad. It's fine for production, as long as you don't have thousands of accounts, which current implementation is not really practical anyway. Basically, if you're comfortable keeping the accounts an a config file, then it's going to be fine.

Besides, even when key look ups vary from key to key, it still doesn't actually leak whether a key is in the map or not. It varies based on whether or not there is a collision; but that doesn't tell you if the key that caused the collision is in the map or not. It won't leak anything about the presence of a particular user name.

THS-on: I agree that this is not that bad, but either we make sure these types of timing attacks do not exist in our implementation, or explicitly acknowledge them that we exclude those. Because otherwise, we will see automated code scans and manual reviews flagging things like this.

setrofim: Sure, but I don't see how variations in key look up can be turned into a timing attack. The variation is due to whether or not there was a collision, which is orthogonal to whether or not that specific key is in the map. The fact that the timing is constant doesn't mean it's exploitable.

THS-on: There is no guarantee in the go implementation that a key miss has the same constant time as a found key. Running a very crude benchmark shows that a miss is likely faster.

setrofim: Ok, I've changed it to loop through all entries.

auth/basic.go
```diff
@@ -103,30 +103,48 @@ func (o *BasicAuthorizer) GetGinHandler(role string) gin.HandlerFunc {
...
 			[]byte(password),
-		); err != nil {
-			o.logger.Debugf("password check failed: %v", err)
+		); err != nil && !haveProblem{
+			o.logger.Warnf("password check failed: %v", err)
+			haveProblem = true
+		}
+
+		if haveProblem {
 			c.Writer.Header().Set("WWW-Authenticate", "Basic realm=veraison")
```
yogeshbdeshpande, line 147: What does the string "realm=veraison", meant to be used for?


I thought the frontend Auth is for the two services, provisioning and verification?

yogeshbdeshpande (approved):
lgtm!, just one minor question

auth/basic.go
setrofim: This indicates the "scope" of protection enforced by this authorizer. Essentially, it indicates a single authentication scheme/database. 

All services (currently provisioning and management) are conceptually part of the same realm.

This is not used for basic auth, as it is internal to services, but set for consistency with Keycloak authentication (which does use realms).

auth/basic.go
```diff
@@ -103,30 +103,48 @@ func (o *BasicAuthorizer) GetGinHandler(role string) gin.HandlerFunc {
...
-		userInfo, ok := o.users[userName]
-		if !ok {
-			c.Writer.Header().Set("WWW-Authenticate", "Basic realm=veraison")
-			ReportProblem(c, http.StatusUnauthorized,
-				"no Basic Authorizaiton given")
-			return
+		var userInfo *basicAuthUser
+		found := false
+		for name, info := range o.users {
+			if name == userName {
```
THS-on: Technically this comparison is not constant time, but I also see no good was of fixing this.

https://pkg.go.dev/crypto/subtle#ConstantTimeCompare would still act differently if the sizes are differently. We can just use add a note that we know and get this merged.

cowbon:
FYI, a similar PR is now opened https://github.com/veraison/ratsd/pull/96 to fix the same issue in ratsd

auth/basic.go
setrofim: Ok, I _think_ I fixed it Using `ConstantTimeCompare()` between the longer string and a buffer of the same length that has the shorter string copied into it.

THS-on (approved):
LGTM! Thanks for fixing this

THS-on:
> FYI, a similar PR is now opened https://github.com/veraison/ratsd/pull/96 to fix the same issue in ratsd
@cowbon can you adopt the same change over in ratsd?

cowbon:
> > FYI, a similar PR is now opened [veraison/ratsd#96](https://github.com/veraison/ratsd/pull/96) to fix the same issue in ratsd
> > @cowbon can you adopt the same change over in ratsd?

Hi @THS-on , yes I've adopted the same change, including your suggestion on the usage of `ConstantTimeCompare()` in https://github.com/veraison/ratsd/pull/96

PR #470, opened 2026-09-16
Get CCA realm attestation working in the end-to-end flow

setrofim:
- Provision realm, as well as platform, endorsements
- Update validity period for the realm endorsements CoRIM
- (not directly related) (re-)enable TLS inside config-docker.yaml


thomas-fossati (approved):
🚢 it!

